PR #1, opened 2023-07-19
Ignore :list, and :require, make some things in declare-rule optional and add some cvc5 rules

hansjoergschurr:
It's all in the title.


src/cmd_parser.cpp
```diff
@@ -110,28 +111,49 @@ bool CmdParser::parseNextCommand()
...
-        d_lex.parseError("Expected args in declare-rule");
+        Token tok = d_lex.peekToken();
+        if (tok == Token::RPAREN)
+        {
+          break;
+        }
+        std::string keyword = d_eparser.parseKeyword();
+        if (keyword == "premises")
+        {
+          premises = d_eparser.parseExprList();
```
ajreynol, line 131: We need to ensure that e.g. `premises` is not set more than once.

src/expr_parser.cpp
```diff
@@ -437,6 +437,18 @@ std::vector<Expr> ExprParser::parseAndBindSortedVarList()
       ss << "Failed to bind symbol " << name;
       d_lex.parseError(ss.str());
     }
+    if (allowList)
```
ajreynol, line 440: I don't like adding this here. I will add support for this separately.

src/cmd_parser.cpp
```diff
@@ -110,28 +111,49 @@ bool CmdParser::parseNextCommand()
...
+          continue;
+        }
+        if (keyword == "conclusion")
+        {
+          conc = d_eparser.parseExpr();
+          continue;
+        }
+        if (keyword == "require")
+        {
+          d_eparser.eatSexpr();
```
ajreynol, line 146: I think we should parse an expression here and type check it.

I will add support for a command `program`.

tests/cvc5-rules.smt2
```diff
@@ -0,0 +1,356 @@
...
+    :require ((C) (= T C RemoveTermFormulas_getAxiomFor t))
+)
+
+; THEORY_LEMMA
+(declare-rule theory_lemma ((F Bool) (tid Int))
+    :conclusion F
+)
+
+; THEORY_REWRITE
+(declare-rule theory_rewrite ((F Bool) (tid Int) (rid Int))
```
ajreynol, line 88: `tid`, `rid` are not `Int`, they should be a auxiliary sort e.g. `(declare-sort TheoryId 0)`.

tests/cvc5-rules.smt2
```diff
@@ -0,0 +1,356 @@
...
+; MACRO_SR_EQ_INTRO
+(declare-rule marco_sr_eq_intro ((T Type) (F Bool :list) (t1 T) (t2 T) (idsar Int :list))
+    :premises (F)
+    :args (t1 idsar) ; We have to merge the three arguments since they are all Int
+    :conclusion t2
+    :require (= T t2 (RwSubstSk t1 F idsar))
+)
+
+; MACRO_SR_PRED_INTRO
+(declare-rule marco_sr_eq_intro ((T Type) (F Bool :list) (R Bool) (idsar Int :list))
```
ajreynol, line 47: `idsar` should not be `Int`. We should make this properly typed e.g. by introducing `(declare-sort MethodId 0)`.

We do need to decide on how to handle default arguments (or perhaps we should not allow them.

ajreynol:
Also I'm going to add a `(include <file>)` command. We should split the signature files into the proper places, e.g. `Ints.sig`, `Builtin.sig`, etc.

PR #5, opened 2023-07-21
Add README

hansjoergschurr:
This adds a README explaining how to compile.

It's mostly taken from LFSC

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,65 @@
+# AletheLF Checker
+
+## An Experimental Proof Checker
+
+## Building the AletheLF checker
+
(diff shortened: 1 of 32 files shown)


README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+cd /path/to/alethelf_checker
+mkdir build
+cd build
+cmake ..
+make
+```
+
+The executable, called `afc`, will be created in the build/src folder.
```
ajreynol, line 19: Actually, let's call the exe `alfc`.

ajreynol (changes requested):

PR #12, opened 2023-08-08
Add tracing

hansjoergschurr:
Currently all tracing streams are enabled when doing a debug build (see README)


ajreynol (changes requested):
Can you delete the things we won't use?  In particular I'd prefer a very basic use of exceptions.

PR #16, opened 2023-08-24
Fix resolution to false.

hansjoergschurr:
This broke with the move towards abstract nil-values.  Before the nil in the case of `or` was the value `false`.  However, with the change resulution to the empty clause would return `alf.nil` instead of `false`.

### DIFF
--- a/proofs/programs/Booleans.smt2
+++ b/proofs/programs/Booleans.smt2
@@ -57,7 +57,7 @@
 (program naryElimOr
     ((t Bool))
     (Bool) Bool
-    (((naryElimOr t) (naryElim or (alf.nil Bool) t)))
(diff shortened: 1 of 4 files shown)


proofs/programs/Booleans.smt2
```diff
@@ -99,7 +99,7 @@
 (program naryElimAnd
     ((t Bool))
     (Bool) Bool
-    (((naryElimAnd t) (naryElim and (alf.nil Bool) t)))
+    (((naryElimAnd t) (naryElim and (alf.nil Bool) true t)))
```
ajreynol, line 102: Do we really need to send `(alf.nil Bool)` to this method?

ajreynol (changes requested):

PR #17, opened 2023-08-24
Produce a proper error message if typechecking at parsing fails

hansjoergschurr:
Before there would be a segmentation fault, since the `getKind()` call would be called on a null pointer.  We check for this now and recover the output of the type checker in the error message.

This also reenables some trace messages

### DIFF
--- a/src/cmd_parser.cpp
+++ b/src/cmd_parser.cpp
@@ -573,15 +573,20 @@ bool CmdParser::parseNextCommand()
         children.push_back(e);
       }
       // compute the type of applying the rule
+      std::stringstream ss;
       Expr concType;
       if (children.size()>1)
       {
         // check type rule for APPLY directly without constructing the app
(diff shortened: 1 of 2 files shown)


src/cmd_parser.cpp
```diff
@@ -573,15 +573,20 @@ bool CmdParser::parseNextCommand()
         children.push_back(e);
       }
       // compute the type of applying the rule
+      std::stringstream ss;
```
ajreynol, line 576: You should not allocate this by default. You should redo the type checking if it fails.

ajreynol: We should just do something like `ExprParser::typeCheckApp` here. I can do this separately if you want.

ajreynol:
I did an alternate fix here: https://github.com/ajreynol/AletheLF-tc/commit/fa5f719ee755a166c7f13327075831d0e70b3613. Does this work?

PR #23, opened 2023-09-21
Handle filepaths without C++ 17 filepath library

hansjoergschurr:
This adds a naive filepath handler that doesn't rely on the C++ filepath API.

Should throw an error when run on Windows.  I think in the future we should detect if the filepath API is available (how?) and use that one in case it is.

### DIFF
--- a/src/compiler.cpp
+++ b/src/compiler.cpp
@@ -6,6 +6,7 @@
 #include "base/check.h"
 #include "base/output.h"
 #include "state.h"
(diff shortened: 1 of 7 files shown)


ajreynol:
Can you make this optionally enabled? So have it optionally use the old implementation, controlled by `#define USE_CPP_FILESYSTEM` (enabled by default).

ajreynol (approved):
Looks good, thanks for the update.

PR #24, opened 2023-10-06
Communicate with oracles by stdin/stdout.

hansjoergschurr:
This changes the oracle handling to use pipes to talk the oracle.  The arguments are written line by line into stdin and stdout is interpreted as one term.

In addition of the arguments the first line written to the oracle is '(',  the second line is the rule name that invoces the oracle, and the last line is ')'.  This allows us to use the same oracle script for multiple rules.  Furthermore, we might later want to have oracles that maintain persistant state. In this case the () help to distinguise different invocations.


ajreynol (changes requested):

ajreynol:
No worries about updating the cvc5 ALF branch, the version is synchronized with the `./contrib/get-alf-checker` script.

ajreynol (changes requested):
Can you add the oracle examples to the CMakeLists? Also I think the oracle_true needs to `echo "true"`.

ajreynol (approved):

PR #35, opened 2024-05-16
Check that the type of the conclusion term is actually "Bool"

hansjoergschurr:
This fixes an issue where some nonsensical proofs would pass checking:
```clojure
(declare-type Pair (Type Type))
(declare-const pair (-> (! Type :var U :implicit) (! Type :var T :implicit) U T (Pair U T)))

(declare-rule partialEva
  ((T Type) (x T))
  :args (x)
  :conclusion x
)

(step c1 :rule partialEva :args ( (pair 4 5) ) )
```

While the rule correctly evaluates here, it doesn't evaluate to a Boolean expression, and checking should fail.

### DIFF
--- a/src/cmd_parser.cpp
+++ b/src/cmd_parser.cpp
@@ -831,6 +831,15 @@ bool CmdParser::parseNextCommand()
         ss << "Non-proof conclusion for rule " << ruleName << ", got " << concType;
         d_lex.parseError(ss.str());
       }
+      // Check that the proved term is actually Bool
+      Expr concTerm = concType[0];
+      Expr concTermType = d_eparser.typeCheck(concTerm);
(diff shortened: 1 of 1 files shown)


ajreynol:
I am waiting to check performance impact of this change.

ajreynol:
As I mentioned, this leads to a roughly 2% downgrade in performance in my testing. This is somewhat acceptable.

However, I think the larger question which @tinelli brought up is whether proof truly needs to be only things that are Bool.

In particular, do we only need a proof of false to justify whether the proof is correct? Or do we need to ensure that all proof steps are for formulas only?   We should discuss this more.

ajreynol:
I think I agree with this, let's merge this for now and look for optimizations if needed later.

PR #37, opened 2024-05-30
Generalize compiler to plugin

ajreynol:
Work towards removing the C++ compiler from the ALF core, and supporting further extensions of the proof checker.


ajreynol:
Let's add further callbacks as necessary in later PRs.

I've added doc in my last commit.

PR #52, opened 2024-06-24
Accept a variant of RARE as an alternative

ajreynol:
This PR essentially makes ethos accept a variant of the syntax for RARE.


ajreynol:
As an update: it is unclear if we will support RARE, or if rewrites will be expressed in native Eunoia moving forward.

PR #66, opened 2024-08-21
Stratify options for smt2/eo, make numeral normalize option

ajreynol:


ajreynol:
The main motivation is to have a way to handle reference checking in logics with reals but not integers.  This permits a way to have the *.smt2 automatically convert the numerals to rationals (`--smt2-normalize-num`) without impacting how Eunoia is parsed.  This then enables faithful reference checking for SMT-LIB compliant parsing for logics like QF_LRA.

There may be further cases where one wants to have 2 different policies for parsing *.eo vs. *.smt2.

PR #72, opened 2024-09-03
cmake: Add option to enable/disable Oracles

daniel-larraz:
This PR introduces a CMake option `ENABLE_ORACLES` and a corresponding compiler option `EO_ORACLES` to control whether Oracles are enabled in the build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
 
     - name: Configure
       if: matrix.build-type == 'debug'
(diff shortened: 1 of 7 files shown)


ajreynol:
@daniel-larraz On second thought, I think we should enable them by default, can you do this?

daniel-larraz:
> On second thought, I think we should enable them by default, can you do this?

@ajreynol  Done

> I am also not a huge fan of removing the source files conditionally from the source tree. I would prefer we just macro-guard the entire file.

@hansjoergschurr  Changed

PR #81, opened 2024-10-01
Fix bug in eo::cons for left associative operators

ajreynol:
The previous code mistakenly would try to append the tail of the cons to the head; instead the head should be appended to the tail in the second argument index.


ajreynol:
The example will be added on https://github.com/cvc5/ethos/pull/83 (that's where i discovered the issue).

ajreynol:
What example do you have in mind?

PR #82, opened 2024-10-01
Add support for define attribute :is_eq

ajreynol:


hansjoergschurr (commented):
Since types can't contain stuck terms anymore, couldn't the same be achieved using `eo::requires` i.e.
```scheme
(declare-const test1 (eo::requires (eo::add 1 1) 2 Bool))
```

ajreynol:
Hm that's true but its a bit of a hack I think...  I'll think about this more, it's not high priority.

hansjoergschurr (approved):

PR #83, opened 2024-10-01
Add support for collapsing variants of nil terminated operators

ajreynol:
Review https://github.com/cvc5/ethos/pull/81 and https://github.com/cvc5/ethos/pull/84 first.


ajreynol:
At the moment we have decided against this approach as, RARE rules will use a "singleton elimination" side condition explicitly.

PR #85, opened 2024-10-26
Check if head is in list form before evaluating eo::list_find

ofecisrael:
For example, before this commit (eo::list_find or (and a b b) a) == -1


ajreynol (approved):
LGTM, thanks for the fix!

PR #93, opened 2024-11-06
ci: Add platform-dependent artifact setup

daniel-larraz:
Licensing information is included with all artifacts. The Windows artifact includes the necessary libraries, which are typically not installed on the system, along with the ethos binary.


daniel-larraz:
> I wonder whether there are licensing issues, given that those dlls are likely covered by certain open source licenses. Maybe we should also copy some COPYING file into the artifact?

Good point. I added licensing information to the artifacts.

PR #99, opened 2024-11-30
Fix out of range error in file extension check

ofecisrael:
This pull request fixes an issue where attempting to check the file extension using substr could cause an out of range error if the file name is shorter than 3 characters.
Added a condition to ensure the file name length is at least 3 before performing the substr operation.
It will prevent potential crashes or undefined behavior when handling short file names.


ajreynol:
Nice catch.

PR #104, opened 2024-12-13
Fix segfault when trace tag is missing after -t option

ofecisrael:
Added error handling to check for a missing trace tag after the -t option. Now displays an appropriate error message instead of causing a segmentation fault.


ajreynol (approved):
Good catch!

PR #117, opened 2025-02-13
Minor changes in user_manual.md

Mallku2:
Small changes to be pushed before they are lost in newer versions of the manual.


user_manual.md
```diff
@@ -235,7 +235,7 @@ In contrast, the example below declares a predicate `=` where the type of the ar
 (declare-const = (-> (! Type :var T :implicit) T T Bool))
 (define P ((x Int) (y Int)) (= x y))
 ```
-
+<!-- should it be "In general, a type argument can be made implicit if..."? -->
```
ajreynol, line 238: Yes, this would apply only to types.

Mallku2: Perfect, changed

user_manual.md
```diff
@@ -236,7 +236,7 @@ In contrast, the example below declares a predicate `=` where the type of the ar
 (define P ((x Int) (y Int)) (= x y))
 ```
 
-In general, an argument can be made implicit if its value can be inferred from the type of later arguments.
+In general, a type argument can be made implicit if its value can be inferred from the type of later arguments.
```
ajreynol, line 239: Actually, sorry, I mispoke, while generally implicit arguments are for types, a non-type can be an implicit argument e.g. if it an argument to a type that is implicit. I would simply omit this change.

ajreynol (changes requested):
One minor comment, else LGTM, thanks for the updates.

user_manual.md
Mallku2: oh!, I see, thanks for the clarification. Already fixed that, sorry for the delay!

ajreynol (approved):
LGTM

Issue #118, opened 2025-02-14
Semantics of `:pairwise` when partial evaluation is involved


Mallku2:
If I define a constant function as:

```
(declare-const func (-> (! Type :var A :implicit) A A Bool) :pairwise and)
```

where `and` is defined as:
```
(declare-const and (-> Bool Bool Bool) :right-assoc-nil true)
```
Then, the expression `(func true)` reduces to `(and)`. In those cases of partial application, I would have expected `(func true)` to not reduce further and stay as `(func true)`. What is the reason behind this semantics?, thanks!

Mallku2:
I wonder if another possibility would be for `(func true)` to reduce to the "neutral" of the operator mentioned in `:pairwise`.

ajreynol:
Good catch, this was unintended. So I think the reasonable semantics here is probably just to throw a parse error, although maybe the neutral element could be used.  Do you have a reason why the latter may be better?

ajreynol:
Btw, `(and)` here is actually `and` applied to 0 arguments, which is a type error. So the error message could be improved, but the behavior is essentially that this is a parse error currently.

Mallku2:
I was thinking about the possibility of using that semantics to define things like the `distinct` operator, which is "variadic" but not in the sense specified by attributes like `:righ-assoc`. That is, `distinct` can be applied to just one parameter, and it yields a boolean, not a function value (contrary to what could happen with a "partial application" of a binary function over just one value, for example).

But, maybe, I am missing some other mechanism in Eunoia to express things like `distinct`. What do you think?

Mallku2:
One thing to add to the previous: I don't want to say that I pretend to define every aspect of the semantics of things like `distinct` just by using a `declare-const` with some special attributes. I understand that, to that end, we have the rules that describe every aspect of the semantics of a given construction, among other things.

The concern here is that, by not implementing the semantics of a variadic operator like `distinct` at the level of its declaration as a constant, and by formalizing a rule like `distinct_elim`, we would be saying that a term like `distinct t` (a function), is "equivalent" to the boolean constant `true`. So I would assume that it is better for `distinct t` to a boolean constant from the point of view of the type system, from the beginning.

ajreynol:
I think I am on board with this change (fwiw, it is a one line change to Ethos to adopt these semantics).  I'll send a PR shortly.

Mallku2:
Awesome, thanks!

PR #120, opened 2025-02-19
Reduce pairwise and chainable for single arguments to neutral element

ajreynol:
Following the discussion on https://github.com/cvc5/ethos/issues/118.


ajreynol:
I believe we are still undecided about this change @hansjoergschurr @Mallku2

PR #154, opened 2025-05-29
Allow optional :signature attribute for program types

ajreynol:


ajreynol:
Should be fixed now, PTAL

PR #161, opened 2025-06-13
Fix list concat with non-atomic null terminator

ajreynol:
Introduced by last commit, adds a regression to ensure this is tested.


ajreynol:
Note that the fix described in this PR actually was included in commit https://github.com/cvc5/ethos/commit/203a25035fd9099f1ffd13a0763dab3d817ff848.

PR #165, opened 2025-07-08
Minor updates

ajreynol:
Taken from smtMeta dev branch.

Some changes are motivated by updates to the plugin interface.

Others eliminate undocumented and unused features.


ajreynol:
Ah, I left out a key line: `eo::self` is no longer bound at global scope (see my last commit).  So, the change to add `eo::self` should be seen as making things more precise (it is only local bound when parsing `declare-consts`).

PR #170, opened 2025-07-22
Minor change to proof rule semantics

ajreynol:
The commit https://github.com/cvc5/ethos/commit/de1a0ece902f553147d902264e1b27f90d895345 updated the type checker for applications of constants to also consider their type.

The above change also was applied to *proof rules*, as one can view proof rules as constants.

However, for simplicity, we can also see proof rules as programs, in which case it is simpler and more intuitive for proof rules to not automatically bind the parameters in types.

One regression had been added demonstrating this feature, we fix it and give an explanation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@ ethos 0.2.1 pre-release
 
 - Adds builtin list operators `eo::list_diff` (difference) and `eo::list_inter` (intersection).
 - Removes support for oracles and the command `declare-oracle-fun`. Custom extensions are now recommended to be added via the plugin feature, which provides an interface for custom evaluation (Plugin::hasEvaluation).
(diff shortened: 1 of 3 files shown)


ajreynol:
Thanks for the response, this is helpful. We should discuss this further with a larger group. Some initial thoughts:

I think there is a case to be made for "proof checking as program evaluation", namely it is more intuitive, given the complexity of the type system. However, I do see your points that there are advantages to "proof checking as type checking". It is more pedantic and indeed it allows proof manipulating programs which is lost in the former viewpoint.

Overall, I think it is fruitful to consider both viewpoints moving forward and attempt to reconcile them. What is interesting is that, at least in practice, the two viewpoints are nearly equivalent (one of the key points seems to be the semantics of eo::typeof vs. the actual type system). Nevertheless, I do think there is a fairly large gap to fully reconcile the two viewpoints, in spite of this being a one line change to ethos.  

I will mark this PR as "draft" for now.

PR #183, opened 2025-10-20
Proof rules as programs

ajreynol:
This changes the internal implementation of ethos so that proof rules are (references to) programs, not constants.

This makes the implementation more uniform and will lead to several forthcoming simplifications.

Furthermore updates a regression which was based off of a BV RARE rule with implicit evaluation in premises, using a similar fix as applied to the one in official Cpc.eo.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,15 @@
 This file contains a summary of important user-visible changes.
 
+ethos 0.2.2 prerelease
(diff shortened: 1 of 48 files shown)


hansjoergschurr (approved):
Can't say that I can fully follow the logic here, but I don't see any obvious errors.

PR #184, opened 2025-10-21
HO applications do not desugar based on the operator

ajreynol:
Also makes a few minor updates to the user manual.

### DIFF
--- a/user_manual.md
+++ b/user_manual.md
@@ -461,6 +461,42 @@ Examples of this desugaring are given below.
 Note that in the case of `(or z)`, no application of `or` is constructed, since only one argument term is given, since it is marked with `:list`.
 In contrast, `(or x)` denotes the `or` whose children are `x` and `false`.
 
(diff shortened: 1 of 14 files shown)


user_manual.md
```diff
@@ -567,6 +574,39 @@ Furthermore, note that a binder also may accept an explicit term as its first ar
 In the above example, `Q4` has `(@cons x)` as its first argument, where `x` was explicitly defined as a variable.
 This means that the definition of `Q4` is also syntactically equivalent to the definition of `Q1` and `Q2`.
 
+#### Further notes on constants with attributes
+
+We have described ways Ethos parses (or _desugars_) applications of the form `(f t1 ... tn)`,
```
hansjoergschurr: Markdown handles underlines (`_foo_`) as italic ( _foo_ ).  This is the same as asterisks: `*foo*`.  I think we should be consistent in the raw text.  For now we use asterisks at most places.  I suggest changing the underlines to asterisks.

PR #188, opened 2025-12-02
Eliminate annotated parameter

ajreynol:
This refactors how parameters in `declare-parameterized-const` are interpreted.

In particular, we determine if the parameter is used in the return type. If yes, we ensure its type is ground. If not, it is treated as an ordinary function argument.

For the former, this disallows a definition like:
```
(declare-parameterized-const id ((T Type :implicit) (x T)) Type)
(declare-parameterized-const bad ((T Type :implicit) (x T)) (-> T (id x)))
```

This furthermore eliminates ANNOT_PARAM, and adds an additional check to ensure that quoted arguments match the type of the provided argument.

Furthermore removes some unused portions of the code, including type check interface for plugins and the `:requires` annotation for parameters.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@ ethos 0.2.3 prerelease
 ethos 0.2.2
 ===========
 
+This release of Ethos is associated with the 1.3.2 release of the SMT solver cvc5.
+
 - The attribute `:signature` to specify the argument and return types of programs is now required.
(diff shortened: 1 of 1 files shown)


hansjoergschurr (approved):
Looks great

hansjoergschurr (approved):
looks great

PR #203, opened 2026-04-01
Fixes for user manual

ajreynol:
Co-Authored-By: ChatGPT 5.4 <noreply@openai.com>


user_manual.md
```diff
@@ -37,28 +37,28 @@ Ethos can be run from the command line via:
 ethos <option>* <file>
 ```
 
-The set of available options `<option>` are given in the appendix. Note the command line interface of `ethos` expects exactly one file (which itself may reference other files via the `include` command as we will see later). The file and options can appear in any order.
+The set of available options `<option>` are given in the appendix. Note that the command line interface of `ethos` accepts at most one file path. If no file path is provided, Ethos reads from stdin. The file and options can appear in any order.
```
hansjoergschurr: This is wrong: ethos only reads from stdin if the input is a pipe.  If you run ethos without an argument you will get an "Error: no input specified.".
The old text was incomplete, because it didn't say that Ethos would read a pipe, but less misleading.  Furthermore, if the pipe-reading functionality is discussed here, we should also say whether that  input is read as a proof or as a signature.

I kindly request to not be assigned to review pull requests that were generated using generative AI in the future.

user_manual.md
```diff
@@ -37,28 +37,29 @@ Ethos can be run from the command line via:
...
+Proof files may either define the symbols they use directly or refer to a Eunoia file that defines them via an `include` command or by using the command line option `--include=X`.
 Complete details on the categories of files accepted by Ethos are described later in this document [here](#full-syntax).
 
 When invoking Ethos on the command line, Ethos will either emit an error message indicating:
 
 - the kind of failure (type checking, proof checking, lexer error)
 - the line and column of the failure
 
-or will print a [successful response](#responses) when it finished parsing all commands in the file or encounters and `exit` command.
+or will print a [successful response](#responses) when it finishes parsing all commands in the file or encounters an `exit` command.
```
Mallku2, line 56: The responses emitted by ethos are the result of parsing + type checking + checking rule applications, right? I would change that to not confuse the user.

user_manual.md
```diff
@@ -501,26 +505,26 @@ in the example above.
...
-In the above example, `(>= x y z w)` is syntax sugar for `(and (>= x y) (>= y z))`,
+In the above example, `(>= x y z)` is syntax sugar for `(and (>= x y) (>= y z))`,
 whereas the term `(>= x y)` is not impacted by the annotation `:chainable` since it has fewer than 3 children.
 
 Note that the type for chainable operators is typically `(-> T T S)` for some types `T` and `S`,
-where the type of its combining operator is `(-> S S S)`, and that operator has been marked as variadic via some attribute (e.g. `:right-assoc`).
+where the type of its combining operator is `(-> S S S)`, and that operator has been marked as variadic via some attribute (e.g. `:right-assoc` or `:right-assoc-nil`).
 
-A chainable operator applied to a single argument reduces to the neutral element of the combining operator.
+A chainable operator applied to a single argument reduces to the neutral element of the combining operator when that operator has a nil terminator.
```
Mallku2: Maybe we could mention that, if the combining operator does not have a nil terminator, the whole partial application of the chainable operator reduces to a partial application of the combining operator.

user_manual.md
```diff
@@ -597,8 +601,8 @@ not desugared further since `xs` is marked `:list`.
 
 In the above example, `forall` is declared as a binder.
 This indicates that the parser (optionally) accepts a variable list as the first argument when parsing applications of `forall` instead of a term.
-In particular, in the last two commands, the parser accepts `(forall ((x Int)) (P x))` for the variable list containing `x`.
-A variable list parsed in this way binds the symbol `x` to a variable of type `Int` when parsing the remaining arguments of `forall`, i.e. its body.
+In particular, in the first three commands, the parser accepts syntax such as `(forall ((x Int)) (P x))` for a variable list.
```
Mallku2: Since all the term `(forall ((x Int)) (P x))` is not only a variable list, and since the commands referred are not correct, I would change this to something like:
```In particular, in the definitions of constants `Q1` and `Q2`, the parser accepted the construction `((x Int))` to denote a variable list containing variable `x`.```

File: user_manual.md [hunk 5]
```diff
@@ -597,8 +601,8 @@ not desugared further since `xs` is marked `:list`.
...
+A variable list parsed in this way binds the corresponding symbol to a variable of type `Int` when parsing the remaining arguments of `forall`, i.e. its body.
```
Mallku2: In this case I would prefer the previous text, where the actual symbol, `x`, was mentioned.

Mallku2 (commented):
I generally agree with the changes introduced by this PR. I only left minor observations.

user_manual.md
ajreynol: This refers to the single response "success" or "incomplete" which is printed after completing the file.

user_manual.md
ajreynol: This results in an error actually. I will mention this.

user_manual.md
ajreynol: Done.

user_manual.md
ajreynol: Done.

user_manual.md
Mallku2: Then, yes, I wouldn't describe that as just "parsing", to avoid confusing the user. The response is the result of the whole process, beginning with parsing, but continuing with type checking, rule checking, etc.

user_manual.md
Mallku2: oh!, actually good to know that we caught it

Mallku2 (approved):

PR #204, opened 2026-04-01
Minor updates in preparation for ethos compilation plugins

ajreynol:
This contains several minor changes from the ethos compilers branch.

In particular:
- opaque applications are no longer curried (reverting a recent change). This simplifies the Ethos compiler to SMT and Lean significantly.
- We now permit programs with empty list of cases. This is to support auto-generated programs that intentionally have no cases (e.g. the desugared $eo_dt_constructors when no user datatypes are defined).
- We no longer silently declare testers and updaters for datatypes, which was undocumented and unused.
- Improvements to printing rationals. This fixes our printing of integral decimals which were incorrectly printing as numerals.

### DIFF
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -1187,7 +1187,8 @@ Expr State::mkApplyAttr(AppInfo* ai,
       }
       else
       {
-        // construct curried APPLY_OPAQUE application.
+        // Note we do not curry APPLY_OPAQUE applications, as they are simpler
+        // to reason about in flattened form.
         std::vector<ExprValue*> ochildren(vchildren.begin(),
                                           vchildren.begin() + 1 + nargs);
         Expr op = Expr(mkExprInternal(Kind::APPLY_OPAQUE, ochildren));
--- a/src/state.cpp
+++ b/src/state.cpp
@@ -1188,12 +1188,9 @@ Expr State::mkApplyAttr(AppInfo* ai,
       else
       {
         // construct curried APPLY_OPAQUE application.
-        ExprValue* curr = vchildren[0];
-        for (size_t i = 1; i < nargs + 1; i++)
-        {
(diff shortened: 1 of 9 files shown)


src/util/rational.cpp
```diff
@@ -44,7 +44,10 @@ Rational Rational::fromDecimal(const std::string& dec) {
 
 bool Rational::isIntegral() const { return mpz_cmp_ui(d_value.get_den_mpz_t(), 1) == 0; }
 
-std::string Rational::toString(int base) const { return d_value.get_str(base); }
+std::string Rational::toString(int base) const
```
hansjoergschurr, line 47: Isn't `base` an odd name her?

src/state.cpp
```diff
@@ -1187,13 +1187,11 @@ Expr State::mkApplyAttr(AppInfo* ai,
...
-        ExprValue* curr = vchildren[0];
-        for (size_t i = 1; i < nargs + 1; i++)
-        {
-          curr = mkExprInternal(Kind::APPLY_OPAQUE, {curr, vchildren[i]});
-        }
-        Expr op = Expr(curr);
+        // Note we do not curry APPLY_OPAQUE applications, as they are simpler
+        // to reason about in flattened form.
+        std::vector<ExprValue*> ochildren(vchildren.begin(),
+                                          vchildren.begin() + 1 + nargs);
```
hansjoergschurr, line 1194: Should we assert that `vchildren.begin() +1 + nargs` is `< vchildren.end()`?

hansjoergschurr (approved):
Looks good some small suggestions

src/state.cpp
ajreynol: Yes

src/util/rational.cpp
ajreynol: `base` refers to the base of the number to print e.g. base 10?

PR #205, opened 2026-04-10
Fix non-ground nil handling for assoc operators having different element/list types

ajreynol:
This fixes an issue with e.g. `:right-assoc-nil` operators whose type is `(-> T U U)` where `T` is not `U`.

In particular, we were not distinguishing element vs. list in the type inference method in computeConstructorTermInternal.

This furthermore cleans up handling of this method to defer to a new `evaluateNil` method where it is specified whether the instantiated type is the element or the list type. This furthermore cleans up the evaluation of eo::nil to use this method.

Also fixes a low-level issue with `:left-assoc-*` operator evaluation.


NEWS.md
```diff
@@ -13,7 +13,8 @@ ethos 0.2.3 prerelease
   treated as a valid code point.
 - Fixes issue in the parser which did not guard for overflow of 32 bit unsigned
   values.
-
+- Fixes an issue where non-ground nil terminators would not be properly computed for operators having types where the element type was different from the return type of the operator.
```
hansjoergschurr:
```suggestion
- Fixes an issue where non-ground nil terminators would not be properly computed for list construction operators with types where the element type is different from the return type of the operator.
```

src/type_checker.cpp
```diff
@@ -1350,24 +1389,29 @@ Expr TypeChecker::evaluateLiteralOpInternal(
   }
   bool isLeft = (ck == Attr::LEFT_ASSOC_NIL || ck == Attr::LEFT_ASSOC_NS_NIL);
   Trace("type_checker_debug") << "EVALUATE-LIT (list) " << k << " " << isLeft << " " << args << std::endl;
-  // infer the nil expression, which depends on the type of args[1]
-  std::vector<Expr> eargs;
-  eargs.emplace_back(args[0]);
+  // infer the nil expression, which may depend on the type of args[1]
   Expr nilExpr;
   if (k==Kind::EVAL_NIL)
```
hansjoergschurr:
```suggestion
  if (k == Kind::EVAL_NIL)
```

src/type_checker.cpp
```diff
@@ -1350,24 +1389,29 @@ Expr TypeChecker::evaluateLiteralOpInternal(
...
-    return computeConstructorTermInternal(ac, eargs);
+    // Special case, eo::nil has the type itself as args[1].
+    return evaluateNil(args[0], ac->d_attrConsTerm.getValue(), isLeft, args[1]);
+  }
+  else if (ac->d_attrConsTerm.getKind() != Kind::PARAMETERIZED)
+  {
+    // If the nil terminator is not parameterized, just take it
+    nilExpr = ac->d_attrConsTerm;
   }
   else if (args.size()>1)
```
hansjoergschurr:
```suggestion
  else if (args.size() > 1)
```

hansjoergschurr (approved):
Some smaller comments about stylistic changes

PR #210, opened 2026-04-30
Simplify type checking for Quote type

ajreynol:
Quote now only takes parameters, this simplifies the type checker based on this.

### DIFF
--- a/src/type_checker.cpp
+++ b/src/type_checker.cpp
@@ -405,8 +405,8 @@ Expr TypeChecker::getTypeAppInternal(std::vector<ExprValue*>& children,
         // we know that x is a parameter of the same type as t. We don't need
         // to match, we simply update the context.
         hdt = hdt->d_children[0];
-        Assert(hdt->getKind()==Kind::PARAM);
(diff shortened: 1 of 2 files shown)


src/type_checker.cpp
```diff
@@ -379,10 +379,8 @@ Expr TypeChecker::getTypeAppInternal(std::vector<ExprValue*>& children,
...
     ExprValue* hdt = hdtypes[i - 1];
-    // if the argument is (Quote t), we match on its argument,
-    // which along with how ctypes[i] is the argument itself, has the effect
-    // of an implicit upcast.
-    bool isQuote = false;
+    // If the argument is (Quote x), then we ensure that x and t have the same
+    // (ground) type. If so, the context is updated where x is mapped to t.
     bool typeSuccess = true;
     ExprValue* child = children[i];
     if (hdt->getKind() == Kind::QUOTE_TYPE)
```
hansjoergschurr, line 386: maybe add an assert here that it's really just a parameter?

hansjoergschurr (approved):

PR #211, opened 2026-05-04
More preparation for Ethos compiler plugins

ajreynol:
Makes the following changes:
- We now permit forward declarations with non-ground type by checking whether the new type is equivalent modulo renaming.  A regression is added demonstrating this.
- getOrSetLiteralTypeRule is made public and renamed to getLiteralTypeRuleMaybeInit.
- 2 new public methods are added which are needed in various plugins:
  (1) getAttributeTerm, which retrieves the term associated with the attribute of a symbol which impacts how it is desugared, and 
  (2) mkRawExpr, which makes an expression without applying desugaring.
- getConstructorKind is renamed to getAttributeKind.

### DIFF
--- a/src/cmd_parser.cpp
+++ b/src/cmd_parser.cpp
@@ -806,7 +806,8 @@ bool CmdParser::parseNextCommand()
         }
         // ensure that its type is compatible with the previous definition
         d_eparser.typeCheckProgramFwdDecl(pprev, progType, name);
-        // we use the old program that was globally bound already
+        // we use the forward declaration of the program that was globally bound
+        // already
         pvar = pprev;
       }
       else
--- a/src/cmd_parser.cpp
+++ b/src/cmd_parser.cpp
@@ -804,8 +804,9 @@ bool CmdParser::parseNextCommand()
         {
           d_lex.parseError("Cannot define program more than once");
         }
-        // it should be a program with the same type
-        d_eparser.typeCheck(pprev, progType);
+        // ensure that its type is compatible with the previous definition
+        d_eparser.typeCheckProgramFwdDecl(pprev, progType, name);
+        // we use the old program that was globally bound already
         pvar = pprev;
       }
       else
--- a/src/expr_parser.cpp
+++ b/src/expr_parser.cpp
@@ -1299,7 +1299,7 @@ void ExprParser::typeCheckProgramFwdDecl(Expr& prevProg,
     }
   }
   std::stringstream ss;
-  ss << "Foward declaration of program " << progName << " had different type.";
+  ss << "Forward declaration of program " << progName << " had different type.";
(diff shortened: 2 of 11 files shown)


src/cmd_parser.cpp
```diff
@@ -804,8 +804,9 @@ bool CmdParser::parseNextCommand()
         {
           d_lex.parseError("Cannot define program more than once");
         }
-        // it should be a program with the same type
-        d_eparser.typeCheck(pprev, progType);
+        // ensure that its type is compatible with the previous definition
+        d_eparser.typeCheckProgramFwdDecl(pprev, progType, name);
+        // we use the old program that was globally bound already
```
hansjoergschurr: maybe "we use the forward definition". It's not clear what "old program" means here.

src/expr_parser.cpp
```diff
@@ -1272,6 +1272,37 @@ void ExprParser::typeCheckProgramPair(Expr& pat,
...
+      ctx[vso[i].getValue()] = vsn[i].getValue();
+    }
+    if (d_state.getTypeChecker().evaluate(prevType.getValue(), ctx) == newType)
+    {
+      // same after renaming, return
+      return;
+    }
+  }
+  std::stringstream ss;
+  ss << "Foward declaration of program " << progName << " had different type.";
```
hansjoergschurr:
```suggestion
  ss << "Forward declaration of program " << progName << " had different type.";
```

src/type_checker.h
```diff
@@ -45,6 +45,12 @@ class TypeChecker
   static bool checkArity(Kind k, size_t nargs, std::ostream* out = nullptr);
   /** Set type rule for literal kind k to t */
   void setLiteralTypeRule(Kind k, const Expr& t);
+  /**
+   * Get or set type rule (to default) for literal kind k. The argument
+   * self is the expression to instantiate eo::self with, if applicable,
+   * otherwise eo::? is used.
+   */
+  Expr getOrSetLiteralTypeRule(Kind k, ExprValue* self = nullptr);
```
hansjoergschurr: It is not super clear to me what this function does.  It seems to set set the type associated with a literal kind (this is the rule?), but how can I use it to only get the currently assigned type?  The documentation seems to imply that if I call it with only the kind, then the type gets overwritten with `eo::?`.  I suggest to split this function in two (a getter and a setter) and to polish the comment.

hansjoergschurr (changes requested):

src/type_checker.h
ajreynol: I'll rename to `getLiteralTypeRuleMaybeInit`.

hansjoergschurr: Oh yeah that's cleaner!

hansjoergschurr (approved):

PR #219, opened 2026-08-20
Restore working version of cpp compiler

ajreynol:
The C++ compiler plugin provides a way to preload the result of parsing signature files. 

This is based on an older implementation of this idea, which was stale in the ethos repo. The PR reduces its scope to only compile parsing, not side condition / type rule evaluation.

This required a minor modification to the plugin interface which allows a plugin to preempt the need to including a file, which is what the new plugin after compilation does.

Co-Authored-By: ChatGPT 5.6 <noreply@openai.com>


daniel-larraz:
## Review

Verified each finding against the source, and by building the plugin (`build_custom_ethos.sh`) and running the generator over the repo's `tests/*.eo` signatures.

**Verified correct:** the `Plugin::includeFile` bool-return protocol is a clean design — `src/plugin.h:53-68` documents it well, and `src/state.cpp:322-334` correctly restores `d_inputFile` on the early-return path. The `expr_trie.h` use-after-free fix is right. Build is warning-clean under `-Wall`; the CI happy-path test passes.

### Findings

**1. `Kind::VARIABLE` is not a symbol — aborts the generator on 20 of 179 test signatures** — `plugins/cpp_compiler/compiler.cpp:27`

`isNamedExpr` includes `Kind::VARIABLE`, but per `src/state.cpp:1388-1396` variables are deliberately *not* atomic — `getBoundVar` builds a two-child `(eo::var <string> <type>)` via `mkExpr`. So `writeExpr` takes the `mkSymbol` branch at `compiler.cpp:329-334`, where `value->asLiteral()` is null and `lookupType` returns null. Any signature using `:binder` or `eo::var` aborts (`binder-ex`, `skolemize-v1/v2`, `pf-quant`, `eo-definitions`, `par-dt`, `simul-overload`, …). `VARIABLE` should fall through to the ordinary `mkExpr` child-recursion path.

**2. `defineProgram` skipped inside scopes, but programs are global state** — `plugins/cpp_compiler/compiler.cpp:201`

The `d_nscopes > 0` guard is correct for `bind` (scoped params), but wrong for `defineProgram`. `State::mkDisambiguatedType` (`src/state.cpp:916-941`) defines `$eo_disamb_type_*` while a `par` / `declare-parameterized-const` scope is open (`src/cmd_parser.cpp:251`), so the generated code declares the constructor without defining its type program. Reproduced at runtime: `No program defined for $eo_disamb_type_nil`, and a proof the stock binary accepts gets rejected.

**3. Moving `setPlugin` above the include loop breaks `--reference`** — `src/main.cpp:169`

`s.setPlugin(plugin)` moved from *after* the command-line include loop to *before* it. That loop passes `!isInclude` as `isReference`, and `src/state.cpp:325` has a pre-existing `Assert(!isReference)` guarding the plugin call. So `--reference=X` now trips `Check failure !isReference` in debug builds (reproduced). The plugin genuinely needs to be installed before includes for this feature to work, so the fix is probably to relax that assert — both plugins already return `false` for references — rather than move the call back.

**4. Hexadecimal literals lose their bit-width** — `plugins/cpp_compiler/compiler.cpp:323`

Round-tripping through `Literal::toString()` is lossy for hex. `BitVector::toString` (`src/util/bitvector.cpp:36-54`) zero-pads only for base 2, while the string constructor (`src/util/bitvector.cpp:25`) sets `d_size = num.size() * 4`. So an 8-bit `#x00` is emitted as `mkLiteral(Kind::HEXADECIMAL, "0")` and comes back 4-bit — the term no longer compares equal. Emit the padded string, or serialize the width explicitly.

**5. Signature at a different path falls through to double-parsing** — `plugins/cpp_compiler/compiler.cpp:114`

The path-tied limitation is documented in the README, but the failure isn't graceful: a signature at a path other than the generation-time one re-parses on top of the already-reconstructed state and dies with a confusing `cannot set type rule for kind NUMERAL...`. Worth a content hash or at least an explicit diagnostic.

**6. Dead, uncompilable test code** — `plugins/cpp_compiler/test/replay.cpp:64`

`ETHOS_CPP_COMPILER_TEST_SIGNATURE` is defined nowhere in the tree, and neither `test/replay.cpp` nor `test/generate.cpp` is referenced by any CMake target — `plugins/cpp_compiler/CMakeLists.txt` doesn't mention them, and the root `plugins-compile-check` excludes the directory. Either wire them up or drop them.

---

Findings 1 and 2 are the blockers — together they mean the plugin only works on the subset of signatures that avoid binders and ambiguous datatype constructors, which is a narrower guarantee than the PR description implies.

daniel-larraz (changes requested):

daniel-larraz:
## Re-review of `7908ce94`

Verified every fix by building and running, not just by reading the diff.

### All 6 findings addressed

| # | Finding | Fix | Verification |
|---|---|---|---|
| 1 | `VARIABLE` treated as a symbol | dropped from `isNamedExpr`; now emits `mkRawExpr(Kind::VARIABLE, {…})` | generator now succeeds on **177/179** `tests/*.eo` (was 159). The 2 remaining are negative tests that stock `ethos` rejects identically (`eo-definitions.eo`, `parser-overflow-declare-sort.eo`). `pf-quant.eo` checks `correct` through a custom binary. |
| 2 | `defineProgram` skipped inside scopes | `d_nscopes` guard removed from `defineProgram`, kept on `bind` | built a custom binary for a `par`-datatype signature; `(as nil (List Int))` now resolves and the proof checks `correct`. |
| 3 | `--reference` tripped `Assert(!isReference)` | both asserts removed from `src/state.cpp` | debug build (`EO_ASSERTIONS`) with the plugin installed now matches stock output exactly on `--include=… --reference=…`. |
| 4 | hexadecimal literals lost bit-width | `literalText()` zero-pads to `(size + 3) / 4` digits | generated source emits `mkLiteral(Kind::HEXADECIMAL, "00")`; replay test asserts `getSize() == 8`. |
| 5 | wrong-path signature double-parsed | matching switched from path to FNV-1a content hash, plus a diagnostic | better than what I suggested — a relocated signature now matches by content and works. A stale signature gives a clear "regenerate the custom ethos binary" message. |
| 6 | dead test code | `test` CMake mode, `cpp-compiler-roundtrip` target + `ctest`, wired into CI | builds warning-clean under `-Werror`; `ctest` passes. The test genuinely regression-guards findings 1, 2 and 4 (exits 10/11/12). |

The CMake refactor into a shared `ethos-cpp-core` static library is a nice side benefit — the three modes no longer each recompile all of `src/`.

### One new issue introduced by the content-hash fix

`plugins/cpp_compiler/compiler.cpp:158` — the mismatch diagnostic's third condition is `requested.filename() == compiled.filename()`. Unlike the two path checks beside it, that one is a heuristic, and it turns a legitimate include into a fatal error:

```
# custom binary compiled for  dt/dtsig.eo
$ stock  ethos other/proof2.eo   ->  correct
$ custom ethos other/proof2.eo   ->  EO_FATAL: signature other/dtsig.eo does not match
                                     the compiled signature dt/dtsig.eo; regenerate...
```

`other/dtsig.eo` here is a completely unrelated file that merely shares a basename. Given how collision-prone signature names are in this ecosystem (`Ints.eo`, `Builtin.eo`, `Nary.eo`, …), this seems likely to bite in practice.

The two exact checks (`==` and `std::filesystem::equivalent`) are precise and already cover the real "user edited the signature in place" case that the diagnostic is for. Dropping the filename fallback fixes the false positive with no loss of coverage.

daniel-larraz (approved):

PR #220, opened 2026-08-20
Add shared plugin meta-reduction support

ajreynol:
Add the common MetaKind classification and parse-definition naming helpers, together with MetaReducePlugin, the common base used by the SMT and Lean emitters in the forthcoming Eunoia compiler toolchain.

This builds on the generic StdPlugin already present on main without modifying it. Backend configuration remains private to each owning implementation.

### DIFF
--- a/plugins/meta_reduce_plugin.cpp
+++ b/plugins/meta_reduce_plugin.cpp
@@ -1,0 +1,302 @@
+/******************************************************************************
+ * This file is part of the ethos project.
+ *
+ * Copyright (c) 2023-2024 by the authors listed in the file AUTHORS
+ * in the top-level source directory and their institutional affiliations.
+ * All rights reserved.  See the file COPYING in the top-level source
+ * directory for licensing information.
+ ******************************************************************************/
+
+#include "meta_reduce_plugin.h"
+
+#include <fstream>
+#include <sstream>
+
+#include "literal.h"
+
+namespace ethos {
+
+MetaReducePlugin::MetaReducePlugin(State& s, ConjectureType conjType)
+    : StdPlugin(s), d_conjectureType(conjType)
+{
+  initializeCommonMetaKinds();
+}
+
+MetaReducePlugin::~MetaReducePlugin() {}
+
+ConjectureType MetaReducePlugin::optionSmtMetaConjectureType() const
+{
+  return d_conjectureType;
+}
+
+void MetaReducePlugin::initializeCommonMetaKinds()
+{
+  // Note that the types of the deep embedding whose constructors are static
+  // in the backend templates are not listed here; they are declared via the
+  // $native_embed_* symbols in the Eunoia templates, which carry their
+  // embedded names (see getTypeMetaKindFor and getEmbedTypeName).
+  d_typeToMetaKind["$smt_Term"] = MetaKind::SMT;
+  d_typeToMetaKind["$smt_Type"] = MetaKind::SMT_TYPE;
+  d_typeToMetaKind["$smt_Value"] = MetaKind::SMT_VALUE;
+  d_typeToMetaKind["$smt_Map"] = MetaKind::SMT_MAP;
+  d_typeToMetaKind["$smt_Seq"] = MetaKind::SMT_SEQ;
+  d_typeToMetaKind["$native_BuiltinType"] = MetaKind::SMT_BUILTIN;
+
+  d_prefixToMetaKind["sm"] = MetaKind::SMT;
+  d_prefixToMetaKind["tsm"] = MetaKind::SMT_TYPE;
+  d_prefixToMetaKind["vsm"] = MetaKind::SMT_VALUE;
+  d_prefixToMetaKind["msm"] = MetaKind::SMT_MAP;
+  d_prefixToMetaKind["ssm"] = MetaKind::SMT_SEQ;
+}
+
+void MetaReducePlugin::bind(const std::string&, const Expr& e)
+{
+  if (e.getKind() != Kind::CONST)
+  {
+    return;
+  }
+  finalizeDecl(e);
+}
+
+std::string MetaReducePlugin::getName(const Expr& e)
+{
+  std::stringstream ss;
+  if (e.getNumChildren() == 0)
+  {
+    ss << e;
+  }
+  return ss.str();
+}
+
+bool MetaReducePlugin::isEmbedCons(const Expr& e)
+{
+  std::string sname = getName(e);
+  return (sname.compare(0, 5, "$emb_") == 0);
+}
+
+bool MetaReducePlugin::isSmtApplyApp(const Expr& oApp)
+{
+  if (oApp.getKind() != Kind::APPLY_OPAQUE || oApp.getNumChildren() <= 1
+      || oApp[1].getKind() != Kind::STRING)
+  {
+    return false;
+  }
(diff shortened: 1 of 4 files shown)


daniel-larraz:
## Review

Nice, well-scoped addition — nothing on `main` is touched, the CMake glob in `CMakeLists.txt:71` picks the new sources up automatically, and `make plugins-compile-check` builds clean. The doc comments on `MetaKind` and `getParseDefPrefix()` are genuinely useful.

Findings below, most significant first. Since `MetaReducePlugin` is abstract and no concrete backend exists on this branch yet, I verified the first four with a probe harness that subclasses it and re-exposes the protected helpers, linked against `src/` + `plugins/` **without** `-DEO_ASSERTIONS` (matching the repo's Release configuration).

---

### 1. `eo::requires` is not stripped when descending function ranges

`meta_reduce_plugin.cpp:191-194` (`getEmbedTypeApp`) and `meta_reduce_plugin.cpp:131-135` (`getTypeMetaKindFor`) both walk to a function's range with a bare `t[t.getNumChildren() - 1]`. In ethos a function-type range can be wrapped in `Kind::EVAL_REQUIRES` — `State::mkRequires` (`state.cpp:570`) builds exactly that, and it is common in signatures.

The codebase's own range-walker already handles this: `Expr::getFunctionType()` has an explicit inner loop `while (et.getKind()==Kind::EVAL_REQUIRES) et = et[2];` (`expr.cpp:186-189`).

Reproduced:

```
guarded  = (-> T (eo::requires T U ($native_embed_smt "SmtRegLan")))

getEmbedTypeApp(guarded) null?                        YES  <-- miss
getTypeMetaKindFor(guarded, followRange=true)       = EUNOIA   (expected SMT_EMBED)
control, identical type without the requires        = SMT_EMBED
```

So a constructor whose range is guarded by `eo::requires` is silently misclassified as `EUNOIA` instead of its embed kind. Both loops should strip `EVAL_REQUIRES` (child index 2).

*(Note for anyone re-running this: the requires must be non-ground, e.g. over `PARAM`s. With ground arguments `mkExpr` discharges it and the test is vacuous.)*

### 2. `getEmbedTypeName` reads an inactive union member

`Literal` is a genuine `union` (`literal.h:27-35`), and `meta_reduce_plugin.cpp:212` reads `d_str` unconditionally. `getEmbedTypeApp` matches `$native_embed_` by prefix only and never checks `getNumChildren() == 2` or that `t[1]` is a `STRING` — unlike `isSmtApplyApp`, which checks both. So `getEmbedTypeApp` happily accepts an app whose argument is a `NUMERAL` (active member `d_int`) or a `CONST` (active member `d_sym`), and `getEmbedTypeName` then reads it as `d_str`.

The `Assert` at `meta_reduce_plugin.cpp:208-210` does guard this, but `EO_ASSERTIONS` is only defined for `Debug` (`CMakeLists.txt:45-47`), so it vanishes in Release. Segfault under gdb, assertions off:

```
#0  ethos::String::toString(bool) const at src/util/string.cpp:203
#1  ethos::MetaReducePlugin::getEmbedTypeName(Expr const&) at plugins/meta_reduce_plugin.cpp:212
```

Suggest tightening `getEmbedTypeApp` to match `isSmtApplyApp`'s shape check, and/or making this an `EO_FATAL`.

### 3. `getName` returns the printed form, not the symbol name

`meta_reduce_plugin.cpp:61-69` streams the `Expr`, which for symbols routes through `Expr::quoteSymbol` (`expr.cpp:205-238`). Any symbol outside the SMT-LIB simple-symbol charset — or one starting with a digit — comes back wrapped in vertical bars:

```
  raw=[$emb_sm.foo]  getName=[$emb_sm.foo]  getSymbol=[$emb_sm.foo]
  raw=[foo bar]      getName=[|foo bar|]    getSymbol=[foo bar]
  raw=[2x]           getName=[|2x|]         getSymbol=[2x]
  raw=[bv[32]]       getName=[|bv[32]|]     getSymbol=[bv[32]]
```

The prefix tests (`$emb_`, `$native_*`, `@@`, `$eo_`) are unaffected, since `$`, `@`, `_` and `.` are all in the permitted set. The damage is in the fallback at `meta_reduce_plugin.cpp:184-185`, where `cname = sname` is handed to the backends as the name to emit: an ordinary user-declared symbol needing quoting reaches the embedding as `|foo bar|` rather than `foo bar`.

This is reachable from real input — `(declare-const |foo bar| Type)` is accepted, and the parser strips the bars before binding (`expr_parser.cpp:968-973`), so the bound name is `foo bar` and the bars are re-added only by the printer. `Expr::getSymbol()` (`expr.cpp:161-169`) returns the raw string and looks like what's wanted here. One caveat: it returns `""` for non-literal atoms such as `Kind::TYPE`, where the stream form gives `"Type"` — worth checking no `d_typeToMetaKind` lookup depends on that.

### 4. `metaKindToPrefix` is not the inverse of `d_prefixToMetaKind`

`initializeCommonMetaKinds` registers `"msm" -> SMT_MAP` and `"ssm" -> SMT_SEQ` (`meta_reduce_plugin.cpp:47-48`), but `utils.cpp:42-55` has no case for either:

```
  EUNOIA      -> [eo.]                       SMT_MAP     -> [?MetaKindPrefix_SMT_MAP]
  SMT         -> [sm.]                       SMT_SEQ     -> [?MetaKindPrefix_SMT_SEQ]
  SMT_TYPE    -> [tsm.]                      SMT_BUILTIN -> [?]
  SMT_VALUE   -> [vsm.]
```

A backend that round-trips a map/seq constructor name through `metaKindToPrefix` will bake that placeholder into generated output instead of failing loudly. Nothing calls it yet, so this is latent — but the asymmetry is entirely within the two files added here. Adding the `msm.`/`ssm.` cases plus an `EO_FATAL` default (as `literalKindToString` does at `std_plugin.cpp:263`) would make it unrepresentable.

---

### Minor

- **Dead member**: `Expr d_null;` at `meta_reduce_plugin.h:135` is never read anywhere in `plugins/` (only `cpp_compiler` has its own). `getEmbedTypeApp` already returns `Expr()` directly.
- **Naming**: `optionSmtMetaConjectureType()` (`meta_reduce_plugin.h:53`) reads as an option lookup in an SMT-specific class, but it's a plain getter for `d_conjectureType` on the backend-neutral base. `getConjectureType()` matches what it does.
- **Silent template no-op**: in the `replAll == false` branch, `StdPlugin::replace` (`std_plugin.cpp:278-285`) does nothing when the tag is absent, so a mistyped tag yields an unrendered template with no diagnostic. `emitResourceFile` also doesn't check `out` for write errors after `out << rendered` (`meta_reduce_plugin.cpp:299`).
- **Header guard convention**: `PLUGIN_UTILS_H` / `PLUGIN_META_REDUCE_PLUGIN_H` vs. the neighbouring `STD_PLUGIN_H` — worth picking one within `plugins/`.
- `prefixToMetaKind`'s default argument `MetaKind::EUNOIA` is never used; the only call site passes `NONE` explicitly.

### Checked and fine

No defect in `buildLambdaDefineProgram`: `cmd_parser.cpp:652-656` only builds a `LAMBDA` when `vars.size() > 0`, so its `Assert(e[0].getNumChildren() != 0)` — and the matching one in `mkProgramType` — holds even with assertions compiled out.

daniel-larraz (changes requested):

daniel-larraz (approved):

PR #221, opened 2026-08-20
Add linear-pattern transformation for Eunoia programs

ajreynol:
Eunoia permits patterns with duplicated variables while Lean does not.

This introduces a transformation from Eunoia to Eunoia that eliminates such patterns, which will be used in the compiler to Lean.

### DIFF
--- a/plugins/linear_patterns/linear_patterns.cpp
+++ b/plugins/linear_patterns/linear_patterns.cpp
@@ -10,7 +10,6 @@
 #include "linear_patterns.h"
 
 #include <algorithm>
-#include <set>
 #include <sstream>
 
 #include "state.h"
@@ -35,16 +34,18 @@ std::vector<std::pair<Expr, Expr>> LinearPattern::linearize(State& s,
     std::pair<Expr, Expr> lpat = linearizePattern(s, pat);
     if (lpat.second.isNull())
     {
-      currCases.push_back(progDef[i]);
+      currCases.push_back(
+          s.mkPair(mkCasePattern(s, currProg, pat), progDef[i][1]));
       continue;
     }
     if (i + 1 == ncases)
     {
       // as an optimization, just do a requires if we are the last case
       Expr ctrue = s.mkTrue();
-      Expr ret =
+      Expr guardedRet =
           s.mkExpr(Kind::EVAL_REQUIRES, {lpat.second, ctrue, progDef[i][1]});
-      currCases.push_back(s.mkPair(lpat.first, ret));
+      currCases.push_back(
+          s.mkPair(mkCasePattern(s, currProg, lpat.first), guardedRet));
       continue;
     }
     // make a new copy of the program
@@ -69,7 +70,7 @@ std::vector<std::pair<Expr, Expr>> LinearPattern::linearize(State& s,
     Expr newApp = s.mkExpr(Kind::APPLY, newappc);
     Expr retLin =
         s.mkExpr(Kind::EVAL_IF_THEN_ELSE, {lpat.second, progDef[i][1], newApp});
-    Expr linCase = s.mkPair(lpat.first, retLin);
+    Expr linCase = s.mkPair(mkCasePattern(s, currProg, lpat.first), retLin);
     currCases.push_back(linCase);
     // only needs a default if the linearized case was not already fully general
     if (!wasDefault)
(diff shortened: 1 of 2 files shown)


daniel-larraz:
## Review

I compiled the plugin (`make plugins-compile-check` is clean) and, to check behavior rather than just read it, linked `linear_patterns.cpp` into a throwaway driver against `src/` so I could linearize real `.eo` programs, install the result back into the state, and evaluate calls before and after.

The split/continuation design is sound — I checked it against several program shapes and it preserves fall-through semantics. But one issue breaks the main intended use case.

### Blocking: pattern reconstruction re-applies constructor desugaring

`linearizeRec` rebuilds a changed pattern node with `s.mkExpr(pat.getKind(), nchildren)` (linear_patterns.cpp:150). For `Kind::APPLY`, `State::mkExpr` consults `getAppInfo` on the head and re-runs the `:right-assoc-nil` / `:chainable` / constructor-term desugaring the parser already performed. Since `pat` is already in desugared internal form, the second pass corrupts it.

Given:

```
(declare-const and (-> Bool Bool Bool) :right-assoc-nil true)
(program $collapse ((x Bool) (y Bool)) :signature (Bool) Bool
  ( (($collapse (and x x)) x)
    (($collapse (and x y)) (and x y)) ))
```

`linearize` returns (untouched case shown first for comparison):

```
(program $eo.l.1.$collapse
   (($collapse (_ (and x) (_ (and y) true)))  ...)                   ; parser form, correct
)
(program $collapse
   (($collapse (_ (and x) (_ (_ (and $eo.lv.x.2) true) true))) ...)  ; rebuilt, wrong
   (($collapse $eo.dv.1) ($eo.l.1.$collapse $eo.dv.1))
)
```

The inner partial application `(and x)` went back through `mkApplyAttr`, which appended a second nil terminator: `(_ (and $eo.lv.x.2) true)` where the parser produces bare `(and $eo.lv.x.2)`.

Evaluating before and after installing the linearized programs confirms the consequence:

```
########## and, :right-assoc-nil ##########
BEFORE ($nary (and a a)) = a
AFTER  ($entry (and a a)) = (g a a)      <-- wrong, falls through to case 2
BEFORE ($nary (and a b)) = (g a b)
AFTER  ($entry (and a b)) = (g a b)      <-- unchanged, correct

########## plain g (control) ##########
BEFORE ($plain (g a a)) = a
AFTER  ($entry (g a a)) = a              <-- correct
BEFORE ($plain (g a b)) = (and a b)
AFTER  ($entry (g a b)) = (and a b)      <-- correct
```

(`$entry` is a fresh symbol holding the linearized entry point; `defineProgram` refuses to re-mark the original, and case-pattern heads are ignored by `evaluateProgramInternal`, so this is behaviorally equivalent.)

So the transformation is semantics-preserving for a plain operator and silently changes the result for a `:right-assoc-nil` one. `declare-parameterized-const` heads round-trip fine, so the trigger is specifically the constructor attribute.

This is the common case rather than an edge one: `and`, `or`, `+`, `=` and friends in the cvc5 signatures are all `:right-assoc-nil` or `:chainable`, and those are exactly the operators non-linear patterns appear under.

The fix is to reconstruct without desugaring. `State::mkExprInternal` / `mkApplyInternal` are what `TypeChecker::evaluate` uses for the same job (type_checker.cpp:789), so this probably means exposing a raw application builder or befriending `LinearPattern`.

### The ordering guarantee does not hold for recursive programs

The header says the result is ordered "so that a program appears after every continuation it calls, which lets a consumer emit them in order without forward declarations." That holds only if no case body calls the original program. For:

```
(program $rec ((x Bool) (y Bool) (z Bool)) :signature (Bool) Bool
  ( (($rec (g x x)) x)
    (($rec (g x y)) ($rec y))
    (($rec z) z) ))
```

the output is `$eo.l.1.$rec` first (its body calls `$rec`), then `$rec` (its body calls `$eo.l.1.$rec`). The group is genuinely mutually recursive, so no ordering avoids a forward reference. Since Eunoia programs are usually recursive, the doc should say the group may be mutually recursive and that the consumer has to handle it, rather than promising a topological order.

### Continuation programs end up with mixed application heads

The doc notes that carried-over cases keep the original head. But the catch-all built at linear_patterns.cpp:76-80 uses `currProg`, so after two splits a single program contains both conventions:

```
(program $eo.l.1.$two
   (($two (or x ...)) ...)                              ; original head
   (($eo.l.1.$two $eo.dv.1) ($eo.l.2.$two $eo.dv.1)))   ; owning-program head
```

Not a bug under the stated "heads are ignored" contract, but it is a trap for any consumer that does read the head (to detect self-recursion, say). Picking one convention and applying it uniformly seems cheap.

### Smaller things

- linear_patterns.cpp:45: `Expr ret` shadows the function's result vector `ret`.
- linear_patterns.cpp:13: `<set>` is unused.
- No tests. A single round-trip test — linearize a program over a `:right-assoc-nil` operator and compare the rebuilt pattern against the parser's form for the same surface syntax — would have caught the blocking issue directly.

### Checked and fine

Verified by running:

- `mkExpr(EVAL_AND, conds)` with three or more conditions is safe: `State::mkExpr` left-folds n-ary literal ops, satisfying the arity-2 requirement in `checkArity`. Checked with a four-occurrence pattern.
- `eo::eq` / `eo::requires` / `eo::ite` on `PARAM` arguments are not collapsed at construction time; they appear intact in the output.
- `mkExpr(APPLY, {prog, $eo.dv.1, ...})` does not eagerly run the original program — `($poly $eo.dv.1)` survives even where `$poly` has a fully general last case that would otherwise fire.
- The `wasDefault` optimization is right: `(($dup x x) x)` linearizes to a single fully-general case with no catch-all.

Verified by reading the code, not by experiment:

- `eo::ite` is lazy during evaluation (type_checker.cpp:726-745 pushes only the condition, then only the selected branch), so the else-branch call into the continuation is not evaluated when the guard holds and recursive programs will not diverge. My attempt to demonstrate this end-to-end was inconclusive, because recursion in a continuation calls the original symbol, which my fresh-`$entry` workaround leaves intact.
- Fresh-name collisions within a single emitted program are impossible, since a split terminates its program and so at most one linearized case (and one catch-all) lands in each. That is an argument about the control flow; I did not try to build a counterexample.

daniel-larraz (changes requested):

daniel-larraz (approved):

PR #222, opened 2026-08-20
Optionally require that the last step proves false

ajreynol:
This makes ethos optionally more strict, failing if we did not read a step proving false as the last step.


daniel-larraz:
Built the branch and ran the new tests (all 4 pass), then probed the check's boundaries. The wiring is clean and consistent — option plumbing, `--help`, NEWS, and manual all line up, the check correctly runs before `correct` is printed, and the `isPop ? 1 : 0` level arithmetic is right (the pop happens after `notifyStep`, in `cmd_parser.cpp:961`). Three notes about the guarantee it actually provides, and one about a test.

### 1. The proof file can turn the check off

`user_manual.md:2135` says most CLI options are also settable via `set-option`, and `--require-proof-of-false` inherits that through `state.cpp:65-68`. So a proof can disable it:

```
$ cat t1.eo
(set-option :require-proof-of-false false)
(declare-rule prove ((F Bool)) :premises () :args (F) :conclusion F)
(step final true :rule prove :premises () :args (true))

$ ethos --require-proof-of-false t1.eo
correct     # exit 0
```

For every other option this is harmless, but this one exists specifically to constrain what the proof producer got away with — a proof that can relax its own acceptance criterion gives no guarantee at all. Suggest latching it: once set true from the command line, ignore `set-option` attempts to clear it (warn instead).

### 2. "assumption level zero" is not "no free assumptions"

`(assume ...)` at level zero adds an assumption without pushing a scope (`state.cpp:1414-1423`), and `cmd_parser.cpp:111` explicitly notes this isn't checked. So:

```
(declare-rule id ((F Bool)) :premises (F) :conclusion F)
(assume a false)
(step final false :rule id :premises (a))
```

passes with `--require-proof-of-false` (verified, exit 0). The check is only meaningful when a `--reference` file is also given — that's the path where `addAssumption` (`state.cpp:473-487`) validates level-zero assumptions against the reference asserts. Standalone, it's vacuous.

If the intent is genuine strictness, consider also requiring that level-zero assumptions be reference-validated (or that `d_assumptions` be empty). At minimum the manual sentence at `user_manual.md:2105` should say the guarantee is relative to the level-zero assumptions, which are only constrained when `--reference` is used.

### 3. A trailing unclosed `assume-push` doesn't clear the flag

`(step s false ...)` at level 0 followed by `(assume-push b true)` passes on stdin (exit 0), since the flag is only touched by steps. In file mode `includeFile` catches the unbalanced scope first, so this is stdin-only and low severity — but worth a thought if the above is tightened.

### 4. `require-proof-of-false-nested.eo` doesn't exercise the new check through the normal path

Run as a file argument it dies on the pre-existing error instead:

```
Including file tests/require-proof-of-false-nested.eo did not preserve assumption scope.
```

which is why it needed the `TEST_MODE`/STDIN plumbing in `tests/expect_error.cmake`. A nested test that *closes* its scope hits the new error via the plain file path — verified with:

```
(set-option :require-proof-of-false true)
(declare-rule prove-false () :premises () :args () :conclusion false)
(declare-rule scope ((F Bool) (G Bool)) :assumption F :premises (G) :conclusion F)
(assume-push a true)
(step s false :rule prove-false :premises () :args ())
(step-pop final :rule scope :premises (s))
=> Error: the last proof step did not prove false at assumption level zero.
```

That also tests the more interesting case (the `isPop` arithmetic on a step-pop back to level 0), and lets the whole `expect_error.cmake` diff and the `-DTEST_MODE=${ARGN}` change be dropped. If the cmake change is kept: the non-STDIN `list(APPEND test_command "${TEST_MODE}")` branch is currently dead, and `${test_input_args}` is expanded while unset in the else branch (works, but `set(test_input_args "")` up front is clearer).

### Minor

- The predicate `getAssumptionLevel() == (isPop ? 1 : 0) && result == d_false` is duplicated at `state.cpp:1447` and `state.cpp:1612` — a small private helper would keep the two branches from drifting.
- The comment explaining that arithmetic sits at the top of `notifyStep` (`state.cpp:1435`), above the plugin block and well away from both uses. Moving it to the first use would help.
- In the plugin branch the flag is only assigned when `success` is true, so a failing plugin step leaves a stale value. Unreachable today (a failed step is fatal in `cmd_parser`), but the non-plugin branch assigns unconditionally at its success return — worth making the two branches identical.

### Summary

The mechanism does what it says for the cvc5-with-`--reference` workflow. Findings 1 and 2 are about the standalone case, where the check can be disabled by its own input and is satisfied by assuming `false` outright; I'd address at least #1 before merging, and either fix or document #2. #4 is a straightforward test simplification.

daniel-larraz (changes requested):
Please see review comments.

ajreynol:
I've addressed some of the comments. The others I have not addressed since I think they would make things more complicated. Several issues relate to checking cases when we already fail anyways.

daniel-larraz (approved):

PR #224, opened 2026-08-21
Add the desugar plugin

ajreynol:
This plugin defines the first central step in the forthcoming ethos compiler toolchain.

This plugin performs various roles:
- Desugars symbol attributes, e.g. right-assoc-nil impacts the definition eo::nil, datatypes impacts eo::dt_constructors, etc.
- Defines various eo:: operators in terms of vanilla Eunoia programs.
- Sets up a `native_` naming convention which acts as a way of referencing native (Lean/SMT) methods as placeholders in the Eunoia code.
- Defines the Logos checker template, which defines an executable checker that can be compiled optionally at this step.

More documentation on this phase is forthcoming in followup PRs.


daniel-larraz:
## Code review

Built the plugin (`plugins-compile-check` passes clean) and validated the `.eo` templates by assembling `eo_desugar.eo` with stub substitutions for every `$EO_*$` placeholder and running the real `ethos` binary over it, which reproduced several of the findings below directly.

### Correctness

1. **`plugins/desugar/desugar.cpp:904`** — `finalizeDatatype`'s parametric branch builds `(_ ($eo_List_cons c1) (_ ($eo_List_cons c2) ...))` and never emits the `$eo_List_nil` terminator, so the constructor list ends in a one-argument partial application. Ethos accepts it silently, so consumers that recurse to nil get wrong results.

2. **`plugins/desugar/desugar.cpp:919`** — `finalizeBinder` emits `(eo::list_concat $eo_List_cons vs env)`, but `$eo_List_cons` (`eo_desugar_native.eo:63`) is declared with no assoc-nil attribute, so `eo::list_concat` never evaluates. Verified experimentally: `$eo_is_closed` is permanently stuck for *every* binder. Should be `$eo_list_concat`.

3. **`plugins/desugar/eo_desugar_checker.eo:121`** — `$eo_mk_premise_list` uses `eo::nil`/`eo::cons` on the premise-list constructor, whose attributes this stage deliberately strips, so every `:premise-list` rule in the generated checker is stuck. Verified experimentally. Should be `$eo_nil`/`$eo_cons`.

4. **`plugins/desugar/eo_desugar_native.eo:396`** — `$eo_pow`'s `:signature (T Int) T` references an undeclared `Int`; parsing fails with "Could not find symbol Int" unless the user signature happens to declare that exact name. Should be `$eo_Numeral`.

5. **`plugins/desugar/desugar.cpp:584`** — `mkSymbol(e.getKind(), ss.str(), e)` passes the term as its own type; the `ct` computed on the previous line is discarded. No compiler warning, since `Expr` is non-trivial.

6. **`plugins/desugar/desugar.cpp:560`** — raw `os << e` bypasses overload renaming, so for an overloaded name the `$eo_typeof` case keys on `f` while the declaration was emitted as `$eoo_f.2`. Same pattern at lines 380-381, 417, 863, 898, 910, 919, 1255-1282.

7. **`plugins/desugar/eo_desugar_native.eo:46`** — `$eot_uop1/2/3` declare their operator parameter as `($native_type_0 "UserOp")` while `$emb_UOp1/2/3` expect `"UserOp1"`/`"UserOp2"`/`"UserOp3"`. Not rejected by ethos, so it silently mistypes the deep embedding.

### Robustness / hygiene

8. **`plugins/desugar/desugar_checker.cpp:126`** — the 10-argument / 8-premise template limits are guarded only by `Assert`; in release builds an over-limit rule silently emits `a11`/`n9`, failing later as an opaque parse error.

9. **`plugins/desugar/desugar.cpp:973`** — the template `ifstream` and the output `ofstream` are unchecked; a missing template silently writes a zero-byte generated file while reporting success (unlike `copyResourceToOutput`, which `EO_FATAL`s).

10. **`plugins/desugar/eo_desugar.eo:206`** — `$eo_list_find_rec` binds `z` twice (`(z W)` and `(z X)`); the first is dead and the duplicate registers a stale overload for the name.

Items 2, 3 and 4 are the ones that break at runtime today; 1 is silent wrong behavior. The rest are latent.

daniel-larraz (changes requested):
Please see review comments.

daniel-larraz (approved):

PR #225, opened 2026-08-21
Add trim definitions plugin

ajreynol:
A useful optional utility used in parts of the forthcoming Eunoia compiler toolchain.

### DIFF
--- a/plugins/trim_defs/trim_defs.cpp
+++ b/plugins/trim_defs/trim_defs.cpp
@@ -12,7 +12,6 @@
 #include <algorithm>
 #include <cctype>
 #include <fstream>
-#include <functional>
 #include <iostream>
 #include <memory>
 #include <sstream>
@@ -35,8 +34,20 @@ struct SExpr
   std::vector<SExpr> d_children;
 };
 
+[[noreturn]] void parseError(const std::string& message,
+                             const std::string* commandText = nullptr)
+{
+  if (commandText != nullptr)
+  {
+    EO_FATAL() << "TrimDefs: malformed command `" << *commandText
+               << "`: " << message;
+  }
+  EO_FATAL() << "TrimDefs: malformed input: " << message;
+}
+
 /** Read one token, skipping whitespace and comments. */
-std::string nextToken(std::istream& in)
+std::string nextToken(std::istream& in,
+                      const std::string* commandText = nullptr)
 {
   std::string tok;
   char c;
@@ -86,21 +97,29 @@ std::string nextToken(std::istream& in)
           }
         }
       }
-      Assert(closed);
+      if (!closed)
+      {
+        parseError("unterminated string literal", commandText);
+      }
       break;
     }
     else if (c == '|' && tok.empty())
     {
       tok += c;
+      bool closed = false;
       while (in.get(c))
       {
         tok += c;
         if (c == '|')
         {
+          closed = true;
           break;
         }
       }
-      Assert(!tok.empty() && tok.back() == '|');
+      if (!closed)
+      {
+        parseError("unterminated quoted symbol", commandText);
+      }
       break;
     }
     else
@@ -175,33 +194,83 @@ std::string readFullCommand(std::istream& in)
     }
   }
 
-  Assert(depth == 0 && started && !inString && !inQuotedSymbol);
+  if (depth != 0 || !started || inString || inQuotedSymbol)
+  {
+    parseError("unterminated top-level command");
+  }
   return result;
 }
 
-SExpr parseSExpr(std::istream& in, const std::string& firstToken)
+SExpr parseSExpr(const std::string& commandText)
 {
-  SExpr expr;
-  if (firstToken != "(")
-  {
-    Assert(firstToken != ")");
-    expr.d_isString = !firstToken.empty() && firstToken[0] == '"';
-    expr.d_value = firstToken;
-    return expr;
-  }
-
-  expr.d_isList = true;
+  std::istringstream in(commandText);
+  SExpr root;
+  std::vector<SExpr*> stack;
+  bool started = false;
   while (true)
   {
-    std::string tok = nextToken(in);
-    Assert(!tok.empty());
+    std::string tok = nextToken(in, &commandText);
+    if (tok.empty())
+    {
+      parseError("unexpected end of command", &commandText);
+    }
+    if (tok == "(")
+    {
+      if (!started)
+      {
+        started = true;
+        root.d_isList = true;
+        stack.push_back(&root);
+      }
+      else
+      {
+        if (stack.empty())
+        {
+          parseError("unexpected `(` after the command", &commandText);
+        }
+        stack.back()->d_children.emplace_back();
+        SExpr& child = stack.back()->d_children.back();
+        child.d_isList = true;
+        stack.push_back(&child);
+      }
+      continue;
+    }
     if (tok == ")")
     {
-      break;
+      if (stack.empty())
+      {
+        parseError("unexpected `)`", &commandText);
+      }
+      stack.pop_back();
+      if (stack.empty())
+      {
+        std::string trailing = nextToken(in, &commandText);
+        if (!trailing.empty())
+        {
+          parseError("trailing text after the command", &commandText);
+        }
+        return root;
+      }
+      continue;
     }
-    expr.d_children.push_back(parseSExpr(in, tok));
+    if (!started || stack.empty())
+    {
+      parseError("expected `(` at the start", &commandText);
+    }
+    stack.back()->d_children.emplace_back();
+    SExpr& child = stack.back()->d_children.back();
+    child.d_isString = tok[0] == '"';
+    child.d_value = std::move(tok);
   }
-  return expr;
+}
+
+std::string stripQuotedSymbol(const std::string& symbol)
+{
+  if (symbol.size() >= 2 && symbol.front() == '|' && symbol.back() == '|')
+  {
+    return symbol.substr(1, symbol.size() - 2);
+  }
+  return symbol;
 }
 
 bool getSymbol(const SExpr& expr, std::string& symbol)
@@ -210,17 +279,16 @@ bool getSymbol(const SExpr& expr, std::string& symbol)
   {
     return false;
   }
-  symbol = expr.d_value;
-  if (symbol.size() >= 2 && symbol.front() == '|' && symbol.back() == '|')
-  {
-    symbol = symbol.substr(1, symbol.size() - 2);
-  }
+  symbol = stripQuotedSymbol(expr.d_value);
   return true;
 }
 
 std::string unescapeString(const SExpr& expr)
 {
-  Assert(expr.d_isString && expr.d_value.size() >= 2);
+  if (!expr.d_isString || expr.d_value.size() < 2)
+  {
+    EO_FATAL() << "TrimDefs: invalid internal string token";
+  }
   std::string value;
   for (size_t i = 1, n = expr.d_value.size() - 1; i < n; ++i)
   {
@@ -349,12 +417,16 @@ Command parseCommand(const std::string& s_expr_text)
   Command cmd;
   cmd.d_fullText = s_expr_text;
 
-  std::istringstream in(s_expr_text);
-  std::string first = nextToken(in);
-  Assert(first == "(");
-  SExpr root = parseSExpr(in, first);
-  Assert(root.d_isList && !root.d_children.empty());
-  Assert(getSymbol(root.d_children[0], cmd.d_cmdName));
+  SExpr root = parseSExpr(s_expr_text);
+  if (!root.d_isList || root.d_children.empty())
+  {
+    parseError("expected a non-empty S-expression", &s_expr_text);
+  }
+  bool gotName = getSymbol(root.d_children[0], cmd.d_cmdName);
+  if (!gotName)
+  {
+    parseError("expected a command name", &s_expr_text);
+  }
 
   std::unordered_set<std::string> defined;
   std::unordered_set<std::string> bound;
@@ -496,8 +568,11 @@ void TrimDefs::processCommand(Command&& cmd)
 
 void TrimDefs::resolveDependencies()
 {
-  Assert(d_commands.size() == d_cmdDefinedSyms.size());
-  Assert(d_commands.size() == d_cmdSyms.size());
+  if (d_commands.size() != d_cmdDefinedSyms.size()
+      || d_commands.size() != d_cmdSyms.size())
+  {
+    EO_FATAL() << "TrimDefs: inconsistent command dependency graph";
+  }
   for (size_t cid = 0, ncommands = d_commands.size(); cid < ncommands; ++cid)
   {
     std::unordered_set<size_t>& dependencies = d_cmdSyms[cid];
@@ -519,23 +594,10 @@ void TrimDefs::resolveDependencies()
 std::vector<size_t> TrimDefs::getCommandOrder(
     const std::unordered_set<size_t>& retained) const
 {
-  // A DFS postorder puts each command after the retained definitions of the
-  // symbols it references. Cycles have no topological order; encountering a
-  // command already on the current path breaks the cycle deterministically.
-  std::vector<unsigned char> state(d_commands.size(), 0);
-  std::vector<size_t> order;
-  std::function<void(size_t)> visit = [&](size_t cid) {
-    if (state[cid] == 2)
-    {
-      return;
-    }
-    if (state[cid] == 1)
-    {
-      return;
-    }
-    state[cid] = 1;
-
-    std::vector<size_t> dependencies;
+  std::vector<std::vector<size_t>> dependencies(d_commands.size());
+  for (size_t cid : retained)
+  {
+    std::vector<size_t>& commandDependencies = dependencies[cid];
     for (size_t symbol : d_cmdSyms[cid])
     {
       std::map<size_t, std::unordered_set<size_t>>::const_iterator it =
@@ -548,27 +610,53 @@ std::vector<size_t> TrimDefs::getCommandOrder(
       {
         if (dependency != cid && retained.find(dependency) != retained.end())
         {
-          dependencies.push_back(dependency);
+          commandDependencies.push_back(dependency);
         }
       }
     }
-    std::sort(dependencies.begin(), dependencies.end());
-    dependencies.erase(std::unique(dependencies.begin(), dependencies.end()),
-                       dependencies.end());
-    for (size_t dependency : dependencies)
-    {
-      visit(dependency);
-    }
+    std::sort(commandDependencies.begin(), commandDependencies.end());
+    commandDependencies.erase(
+        std::unique(commandDependencies.begin(), commandDependencies.end()),
+        commandDependencies.end());
+  }
 
-    state[cid] = 2;
-    order.push_back(cid);
-  };
-
+  // An iterative DFS postorder puts each command after the retained
+  // definitions of the symbols it references. Cycles have no topological
+  // order; encountering a command on the current path breaks the cycle
+  // deterministically.
+  std::vector<unsigned char> state(d_commands.size(), 0);
+  std::vector<size_t> order;
   std::vector<size_t> roots(retained.begin(), retained.end());
   std::sort(roots.begin(), roots.end());
-  for (size_t cid : roots)
+  for (size_t root : roots)
   {
-    visit(cid);
+    if (state[root] != 0)
+    {
+      continue;
+    }
+
+    state[root] = 1;
+    std::vector<std::pair<size_t, size_t>> stack;
+    stack.emplace_back(root, 0);
+    while (!stack.empty())
+    {
+      std::pair<size_t, size_t>& frame = stack.back();
+      const size_t cid = frame.first;
+      if (frame.second < dependencies[cid].size())
+      {
+        const size_t dependency = dependencies[cid][frame.second++];
+        if (state[dependency] == 0)
+        {
+          state[dependency] = 1;
+          stack.emplace_back(dependency, 0);
+        }
+        continue;
+      }
+
+      state[cid] = 2;
+      order.push_back(cid);
+      stack.pop_back();
+    }
   }
   return order;
 }
@@ -622,7 +710,8 @@ void TrimDefs::finalize()
   std::vector<size_t> toVisit;
   for (const std::string& dt : d_defTargets)
   {
-    std::map<std::string, size_t>::const_iterator it = d_symToId.find(dt);
+    const std::string target = normalizeSymbol(stripQuotedSymbol(dt));
+    std::map<std::string, size_t>::const_iterator it = d_symToId.find(target);
     if (it == d_symToId.end())
     {
       EO_FATAL() << "Could not find target definition \"" << dt << "\"";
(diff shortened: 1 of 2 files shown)


plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+  cmd.d_fullText = s_expr_text;
+
+  std::istringstream in(s_expr_text);
+  if (nextToken(in) != "(")
+  {
+    Assert(false);
+  }
+
+  cmd.d_cmdName = nextToken(in);
+  cmd.d_symbolName = nextToken(in);
```
daniel-larraz: **`declare-datatypes` parses to garbage.**

For `(declare-datatypes ((Color 0)) (((red) (green))))` the second token is `(`, so `d_symbolName` becomes `"("`. The body loop below then starts at `depth == 0`, descends into the sort-dec list, and breaks at *its* closing `)` — the entire constructor list is never read, so no constructor or selector type is recorded as a dependency.

(The singular `declare-datatype` form happens to work, since its second token really is the datatype name.)

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+  {
+    return;
+  }
+  std::map<std::string, size_t>::iterator its =
+      d_symToId.find(cmd.d_symbolName);
+  size_t id;
+  if (its == d_symToId.end())
+  {
+    ++d_idCounter;
+    d_symToId[cmd.d_symbolName] = d_idCounter;
```
daniel-larraz: **Constructors and selectors are invisible to the dependency graph.**

Only `cmd.d_symbolName` is entered into `d_symToId`. For `(declare-datatype Color ((red) (green)))`, `red` and `green` are never registered, so:

* a rule referencing `red` produces no edge to the datatype declaration, and the declaration is trimmed away; and
* `(echo "trim-defs red")` hits the `EO_FATAL "Could not find target definition"` in `finalize()`.

The same applies to selectors, and to `declare-parameterized-const`-style commands that introduce more than one name.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+      if (visited.find(sym) == visited.end())
+      {
+        keep = false;
+        break;
+      }
+    }
+    if (keep)
+    {
+      cdeps.insert(cid);
+    }
```
daniel-larraz: **The parse-definition retention pass has three separate problems.**

1. It tests membership in `visited`, which is the set of reachable *symbols*, not the set of retained *commands*. Those are different sets.
2. A parse def that gets retained never has its own symbol id inserted into `visited`, so nothing can depend on it.
3. It is a single non-fixpoint pass over `d_parseDefCmds`, so the result depends on insertion order.

Together these mean a `$parse_a` whose body mentions a retained `$parse_b` is always dropped. This needs to iterate to a fixpoint, and mark the retained def's own symbol as available on each round.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+    result += c;
+    if (c == '(')
+    {
+      depth++;
+      started = true;
+    }
+    else if (c == ')')
+    {
+      depth--;
+      if (depth == 0 && started) break;
```
daniel-larraz, line 193: **String literals corrupt `readFullCommand`.**

Neither this function nor `nextToken` tracks whether it is inside a string. A `;` inside a string literal is treated as a comment start, which eats the closing `"` and the closing `)`, so `depth` never returns to 0: the rest of the file is swallowed into a single bogus "command" and the `Assert(depth == 0)` below trips. Unbalanced parens inside a string do the same to the depth counter.

This is more than theoretical here — the plugin's own protocol ships whole s-expressions inside `(echo "trim-defs-cmd ...")`. That only survives today because those parens happen to balance.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+{
+  if (msg.compare(0, 10, "trim-defs ") == 0)
+  {
+    d_defTargets.push_back(msg.substr(10));
+    return false;
+  }
+  if (msg.compare(0, 14, "trim-defs-cmd ") == 0)
+  {
+    std::string msgr = msg.substr(14);
+    Command cmd = parseCommand(msgr);
```
daniel-larraz, line 692: **`trim-defs-cmd` commands are indexed before the file that contains them.**

`echo()` fires while the enclosing file is being parsed, but `finalizeIncludeFile()` only fires once that file *finishes*. So a command injected this way is assigned a lower `cid` than every symbol in its enclosing file, and (see the comment on `processCommand`'s edge resolution) gets no edges to them at all, since they are not yet in `d_symToId`.

Since `finalize()` emits commands sorted by `cid`, the injected command is printed ahead of declarations it references.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+  {
+    // replace eo:: by $eo_
+    std::string su = s;
+    if (su.compare(0, 4, "eo::") == 0)
+    {
+      std::stringstream ssn;
+      ssn << "$eo_" << su.substr(4);
+      su = ssn.str();
+    }
+    its = d_symToId.find(su);
```
daniel-larraz: **Unresolved references are silently dropped.**

An edge is only created when the referenced symbol is *already* in `d_symToId`. Anything parsed later is lost with no diagnostic.

Combined with the `finalizeIncludeFile` ordering, a file included partway through another file both loses its references back to the includer's earlier symbols and sorts ahead of them in the output.

Deferring resolution would fix both: keep the raw names on the edge here, and resolve names to ids once in `finalize()` after everything has been indexed.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+{
+  std::string tok;
+  char c;
+  while (in.get(c))
+  {
+    if (c == ';')
+    {
+      // Skip to end of line
+      while (in.get(c) && c != '\n');
+      continue;
```
daniel-larraz, line 64: **A `;` in the middle of a token merges two tokens.**

The comment skip does `continue` without terminating the token in progress, so `Int;comment\nOther` yields the single token `IntOther` — and both `Int` and `Other` vanish from the dependency set.

This should `break` when `tok` is non-empty, mirroring the whitespace case just below.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+    }
+    if (std::isspace(c))
```
daniel-larraz: `std::isspace` takes an `int` whose value must be representable as `unsigned char` or be `EOF`; passing a plain (signed) `char` is UB for bytes >= 0x80, which show up in UTF-8 comments and echo strings. Cast to `unsigned char`.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+      continue;
+    }
+    ss << d_commands[i];
+    ss << std::endl;
+  }
+
+  // write the trimmed to file
+  std::string outPath = getOutputPath("plugins/trim_defs/trim_gen.eo");
+  std::cout << "Write trim-defs " << outPath << std::endl;
+  std::ofstream out(outPath);
```
daniel-larraz, line 813: The stream open is unchecked, and the `"Write trim-defs ..."` line above prints unconditionally — so a failed write is reported to the user as success. `StdPlugin::copyResourceToOutput` uses `EO_FATAL` in exactly this situation; worth matching that.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+#include "trim_defs.h"
+
+#include <cctype>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <unordered_set>
+#include <vector>
```
daniel-larraz, line 21: `std::sort` (line 308) and `std::unique_ptr` (line 208) are used without including `<algorithm>` and `<memory>`. It builds today only through transitive includes.

plugins/trim_defs/trim_defs.h
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * `TrimDefs` reads signature files as raw top-level S-expressions, builds a
+ * dependency graph from commands to the symbols they reference, and preserves
+ * only commands reachable from one or more `(echo "trim-defs <symbol>")`
+ * targets.  This is primarily used after desugaring and model_smt generation,
+ * where the full generated signature can be much larger than a single VC or
+ * proof-rule workflow needs.
+ *
+ * The pass is deliberately text-preserving: it keeps the original command
+ * bodies and prints reachable commands back out in dependency order instead of
```
daniel-larraz, line 52: The pass prints commands in insertion order (`std::sort` over command ids in `finalize()`), not dependency order — the two coincide only when every definition precedes its uses in the input, which the `finalizeIncludeFile` ordering does not guarantee for mid-file includes.

daniel-larraz (commented):
Review of the trimming plugin. It builds cleanly, but the hand-rolled S-expression parser and the reachability pass have several defects that produce silently wrong output rather than errors.

**Correctness**

1. `declare-datatypes` (plural) parses to garbage — the symbol name comes out as `"("` and the constructor list is skipped entirely.
2. Constructors and selectors are never registered as symbols, so nothing can depend on a datatype declaration and it cannot be used as a trim target.
3. The parse-definition retention pass checks symbol reachability instead of command retention, never marks a retained def's own symbol, and is a single non-fixpoint pass — so chained parse defs are always dropped.
4. The tokenizer does not know about string literals; a `;` or an unbalanced paren inside a string desynchronizes `readFullCommand` for the remainder of the file.
5. `trim-defs-cmd` commands are indexed before `finalizeIncludeFile` indexes their enclosing file, so they get no edges to that file's symbols and sort ahead of them in the output.
6. Dependency edges are only created for symbols already seen; forward and back references are dropped without a diagnostic. Resolving names to ids in `finalize()` instead would fix this and (5).
7. A `;` immediately following a token merges it with the next one.

**Minor**: `std::isspace` on a signed `char`, an unchecked `ofstream` reported as a successful write, missing `<algorithm>`/`<memory>`, and a class doc comment that claims dependency ordering.

Nothing here wires the plugin into a test, which is consistent with the other experimental plugins — but given (1)-(3), a small round-trip over a signature containing a datatype and a couple of chained parse defs would pay for itself.

Generated with [Claude Code](https://claude.com/claude-code)

daniel-larraz (changes requested):
Please see review comments.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,735 @@
...
+{
+  Command cmd;
+  cmd.d_fullText = s_expr_text;
+
+  std::istringstream in(s_expr_text);
+  std::string first = nextToken(in);
+  Assert(first == "(");
+  SExpr root = parseSExpr(in, first);
+  Assert(root.d_isList && !root.d_children.empty());
+  Assert(getSymbol(root.d_children[0], cmd.d_cmdName));
```
daniel-larraz: **Blocking: `Assert` is compiled out in Release, so this side effect is discarded.**

`Assert(cond)` expands to `EO_FATAL_IF(false, ...)` when `EO_ASSERTIONS` is undefined ([`src/base/check.h:138`](https://github.com/cvc5/ethos/blob/main/src/base/check.h#L138)) — the argument does not appear in the expansion at all, so `getSymbol` is never called and `cmd.d_cmdName` stays empty. `CMakeLists.txt` only defines `EO_ASSERTIONS` for `Debug`, so this is every Release build.

Verified against the real header:

```cpp
bool sideEffect(std::string& out) { out = "declare-const"; return true; }
std::string name;
Assert(sideEffect(name));
printf("d_cmdName = [%s]\n", name.c_str());
```
```
$ g++ -std=gnu++17 -Isrc t.cpp src/base/check.cpp src/base/output.cpp && ./a.out
d_cmdName = []                  # Release
$ g++ -std=gnu++17 -DEO_ASSERTIONS -Isrc ... && ./a.out
d_cmdName = [declare-const]     # Debug
```

With `d_cmdName` empty, every dispatch on it in this file silently takes the wrong branch:

* `d_alwaysKeep` (line 415) is never set — `declare-consts` is no longer retained unconditionally.
* `d_isTrimDirective` (line 417) is never set — so `processCommand` no longer skips `(echo "trim-defs-cmd ...")` read from the file text, and it is indexed a *second* time alongside the one injected through the `echo()` callback. This commit also removed the old output-time `(echo "trim-defs` text filter, so trim directives can now be emitted into the trimmed file.
* The `declare-datatype` / `declare-datatypes` branches (lines 361, 373) are never taken, re-introducing exactly the constructor/selector blindness this commit set out to fix.
* `d_parseDefCmds` (line 484) is never populated — every parse definition is dropped.
* `hasSortedVariables` (line 397) returns false, so bound parameters are treated as global references.
* The `include` guard in `processCommand` and the `depends` filter at line 711 never fire.

The plugin compiles without a warning in either mode and fails silently at runtime. Fix by hoisting the call out:

```cpp
bool gotName = getSymbol(root.d_children[0], cmd.d_cmdName);
Assert(gotName);
```

It is worth grepping the rest of the tree for the same pattern — the macro makes it easy to hit.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,735 @@
...
+    expr.d_isString = !firstToken.empty() && firstToken[0] == '"';
+    expr.d_value = firstToken;
+    return expr;
+  }
+
+  expr.d_isList = true;
+  while (true)
+  {
+    std::string tok = nextToken(in);
+    Assert(!tok.empty());
```
daniel-larraz: **Release-only infinite loop on truncated input.**

`nextToken` returns `""` at EOF. In Release this `Assert` is gone, so `parseSExpr(in, "")` falls into the atom branch, returns an empty atom, and the enclosing `while (true)` appends it forever — an unbounded-memory hang rather than a diagnostic.

Commands from `parseCommands` are safe because `readFullCommand` guarantees balance first, but the `echo()` path at line 604 hands raw user text straight to `parseCommand`, so this is reachable from `(echo "trim-defs-cmd (foo")`. Needs a real `if (tok.empty()) break;` (or an error) rather than an assertion.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,735 @@
...
+  if (msg.compare(0, 10, "trim-defs ") == 0)
+  {
+    d_defTargets.push_back(msg.substr(10));
+    return false;
+  }
+  if (msg.compare(0, 14, "trim-defs-cmd ") == 0)
+  {
+    std::string msgr = msg.substr(14);
+    Command cmd = parseCommand(msgr);
+    processCommand(std::move(cmd));
```
daniel-larraz, line 693: **This path bypasses `readFullCommand`, so nothing validates the text before parsing it.**

`parseCommand` relies on `Assert(first == "(")` and `Assert(root.d_isList && !root.d_children.empty())` (lines 354-356) to establish its preconditions. Both vanish in Release, and line 357 then indexes `root.d_children[0]` on a potentially empty vector — an out-of-bounds read on a malformed directive.

Since `msgr` comes from the input file rather than from the plugin's own code, these should be real checks with an `EO_FATAL` naming the offending directive, not assertions.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,735 @@
...
+                  "<symbol> is the name of the "
+                  "symbol to trim with respect to."
+               << std::endl;
+  }
+  resolveDependencies();
+
+  std::vector<size_t> toVisit;
+  for (const std::string& dt : d_defTargets)
+  {
+    std::map<std::string, size_t>::const_iterator it = d_symToId.find(dt);
```
daniel-larraz: Targets are looked up verbatim, while symbols are stored after `getSymbol` strips `|...|` quoting and `resolveDependencies` applies `normalizeSymbol` to references. So `(echo "trim-defs |foo bar|")` and `(echo "trim-defs eo::foo")` both hit `EO_FATAL "Could not find target definition"` even though the symbol is present. Running `dt` through the same stripping and `normalizeSymbol` here would make targets accept the same spellings the body syntax does.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,735 @@
...
+
+std::vector<size_t> TrimDefs::getCommandOrder(
+    const std::unordered_set<size_t>& retained) const
+{
+  // A DFS postorder puts each command after the retained definitions of the
+  // symbols it references. Cycles have no topological order; encountering a
+  // command already on the current path breaks the cycle deterministically.
+  std::vector<unsigned char> state(d_commands.size(), 0);
+  std::vector<size_t> order;
+  std::function<void(size_t)> visit = [&](size_t cid) {
```
daniel-larraz: Minor: this is a recursive DFS, so stack depth is bounded by the longest dependency chain among retained commands. For the generated signatures this pass is aimed at that chain can get long; an explicit stack would remove the failure mode. Same applies to `parseSExpr`, though nesting depth there is naturally much smaller.

daniel-larraz (commented):
Re-reviewed at `22498f17`. All eleven findings from the previous round are genuinely addressed, and the rewrite is a real improvement — replacing the ad-hoc token scanner with a proper `SExpr` parse, deferring symbol resolution to `resolveDependencies()`, running the parse-def retention to a fixpoint over *retained commands* rather than reachable symbols, and topologically ordering the output in `getCommandOrder()`. The `d_isTrimDirective` flag is a much cleaner way to suppress the control commands than the old text prefix match.

**One blocking issue in the new code**, plus three smaller ones that share its root cause.

`Assert(cond)` is a no-op in Release builds — `EO_ASSERTIONS` is only defined for `CMAKE_BUILD_TYPE=Debug`, and the non-assertions expansion drops its argument entirely rather than merely skipping the check. Line 357 puts a side-effecting call inside one:

```cpp
Assert(getSymbol(root.d_children[0], cmd.d_cmdName));
```

so in every Release build `d_cmdName` is the empty string and each of the seven dispatches on it in this file takes the wrong branch — `declare-consts` stops being force-retained, all parse definitions are dropped, datatype constructors go uncollected again, and trim directives get double-indexed and can leak into the output. It compiles warning-free in both modes; I confirmed the behaviour directly against `src/base/check.h`, details inline.

The related three are all cases where `parseCommand` is guarded only by assertions on a path that also accepts input-file text (the `echo` `trim-defs-cmd` handler), giving a Release-mode hang or an out-of-bounds read on malformed input where Debug gives a clean abort.

Also noted inline: target symbols are looked up without the `|...|` stripping and `eo::` normalization applied to every other symbol, and the ordering DFS is recursive.

Generated with [Claude Code](https://claude.com/claude-code)

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,827 @@
...
+      else
+      {
+        if (stack.empty())
+        {
+          parseError("unexpected `(` after the command", &commandText);
+        }
+        stack.back()->d_children.emplace_back();
+        SExpr& child = stack.back()->d_children.back();
+        child.d_isList = true;
+        stack.push_back(&child);
```
daniel-larraz, line 234: Nit, not a bug — I verified this is currently correct, but it is worth a comment or a cheap hardening.

`stack` holds raw `SExpr*` into parent `d_children` vectors, which reallocate on `emplace_back`. That is safe here only because of a non-obvious invariant: a parent's `d_children` grows only while that parent is `stack.back()`, by which point any pointer into it has already been popped. I stress-tested with ASan/UBSan on reallocation-heavy input (40 levels x 60 siblings, and 500-deep nesting) and it is clean.

But the invariant is one edit away from a heap-use-after-free, with no compiler or test signal. Storing indices into the parent instead of pointers, or a short comment stating the invariant, would make it safe to modify later.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,827 @@
...
+    {
+      std::pair<size_t, size_t>& frame = stack.back();
+      const size_t cid = frame.first;
+      if (frame.second < dependencies[cid].size())
+      {
+        const size_t dependency = dependencies[cid][frame.second++];
+        if (state[dependency] == 0)
+        {
+          state[dependency] = 1;
+          stack.emplace_back(dependency, 0);
```
daniel-larraz, line 651: Same class, also currently safe: `frame` is a reference into `stack`, and `stack.emplace_back` below can reallocate. It works because `frame.second++` is sequenced before the `emplace_back` and `frame` is not touched afterwards (`cid` and `dependency` are copies). Re-reading `stack.back()` after the push, or indexing by position, would remove the footgun.

plugins/trim_defs/trim_defs.cpp
```diff
@@ -0,0 +1,827 @@
...
+}
+
+}  // namespace
+
+// Main parser from istream
+void TrimDefs::parseCommands(std::istream& in)
+{
+  while (in)
+  {
+    std::string tok = nextToken(in);
```
daniel-larraz, line 515: Minor: this call and the `parseError` in `readFullCommand` (line 199) pass no context, so a malformed *signature file* reports `TrimDefs: malformed input: unterminated string literal` with no indication of which file or command. `finalizeIncludeFile` already has the path — threading it through would make these as actionable as the `trim-defs-cmd` diagnostics, which read very well.

daniel-larraz (commented):
Re-reviewed at `9018717e`. All five findings from the previous round are fixed, and I verified each one rather than reading for it:

**The blocking `Assert` issue is fully resolved.** Every `Assert` in the file is now a real `EO_FATAL` check — `grep -c 'Assert(' plugins/trim_defs/trim_defs.cpp` returns 0 — so the Debug/Release divergence class is gone at the root rather than patched at the one call site. Hoisting `getSymbol` out into `gotName` is exactly right. I compiled the plugin both with and without `EO_ASSERTIONS` and ran 16 parse cases through each: **behaviour is identical in both modes**, including all seven malformed inputs, which now abort with precise diagnostics.

Sample of the new error output:
```
(foo                       -> malformed command `(foo`: unexpected end of command
()                         -> malformed command `()`: expected a non-empty S-expression
(a) (b)                    -> malformed command `(a) (b)`: trailing text after the command
(echo "abc                 -> malformed command `(echo "abc`: unterminated string literal
abc                        -> malformed command `abc`: expected `(` at the start
(declare-const |abc        -> malformed command `(declare-const |abc`: unterminated quoted symbol
```
The previous infinite loop (`(foo`) and the out-of-bounds read now both terminate immediately, and the `commandText` in each message makes them genuinely diagnosable.

**The iterative `getCommandOrder` is exactly equivalent to the recursive version it replaces.** I transplanted both and differential-tested them over 20,000 random graphs (9,814 acyclic, 9,804 with cycles): zero order mismatches, zero membership errors, zero topological violations on the acyclic cases, ASan/UBSan clean. The cycle-breaking behaviour is preserved, and determinism is too, since the per-command dependency lists are sorted/uniqued and the roots are sorted.

**Parsing spot-checks** all behave correctly:

| input | result |
|---|---|
| `(declare-datatypes ((Color 0)) (((red) (green (nxt Color)))))` | defines `Color red green nxt` |
| `(declare-datatype Lst (par (T) ((nil) (cons (hd T) (tl (Lst T))))))` | defines `Lst nil cons hd tl`; `T` correctly bound, not a reference |
| `(program f ((x Int) (y Bool)) (Int Bool) Int (((f x y) (g x))))` | refs `Int Bool g`; `x`/`y` correctly excluded |
| `(echo "hi ; there ) ((")` | single command, no depth corruption |
| `(declare-const \|weird sym\| (-> \|T A\| Int))` | defines `weird sym`, refs `T A` |

Target normalization now matches the reference side via `normalizeSymbol(stripQuotedSymbol(dt))`.

Nothing blocking left. Three minor notes inline: two latent pointer/reference-invalidation footguns that I confirmed are *currently* correct but are fragile under future edits, and the file-level parse errors not carrying the filename.

Generated with [Claude Code](https://claude.com/claude-code)

daniel-larraz (approved):

PR #227, opened 2026-08-24
Add model smt plugin

ajreynol:
This plugin defines the official semantics for SMT-LIB and the correspondence between CPC terms and SMT-LIB terms.

It contains two Eunoia signatures which currently are hardcoded (they will be compiled from a higher-level source and abstracted further in followup PRs).

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/plugins/model_smt/cpc_defs.eo
+++ b/plugins/model_smt/cpc_defs.eo
@@ -1,0 +1,2091 @@
+; The CPC signature written *directly* as a transformation into the deep
+; embedding.
+;
+; This is what the model-smt stage compiles the CPC half of a model from. For
+; each of the 180 symbols it covers it gives one thing:
+;
+;   $eoc_transform_X       the cases X contributes to $eo_to_smt, i.e.
+;                            T -> $smt_Term for a Eunoia term of any type T;
+;   $eoc_transform_type_X  the cases a *type* constructor X contributes to
+;                            $eo_to_smt_type, i.e. Type -> $smt_Type.
+;
+; It declares no constant: the constants are those of the SMT-LIB signature,
+; see plugins/model_smt/smt_defs.eo. A symbol of CPC is *eliminated* here, in
+; the sense that what it denotes is written in the vocabulary of that file
+; rather than in its own.
+;
+; The stage that reads this file has only to concatenate the cases of every
+; $eoc_transform_X into $eo_to_smt, and to copy everything else through. Note
+; the recursive call is to $eo_to_smt rather than to any $eoc_transform_X: what
+; an argument is, is not known here. A type is transformed by $eo_to_smt_type,
+; which stands to types as $eo_to_smt stands to terms; both are declared by
+; plugins/model_smt/model_smt.eo.
+;
+; A case that walks a structure the desugar stage built rather than a term of
+; the input calls a helper program, e.g. the list a distinctness gathered or
+; the variables a quantifier binds. Each such helper is written out just above
+; the symbol that names it; they live in plugins/model_smt/eo_to_smt_aux.eo
+; today. A symbol comes after the ones its cases name, so that nothing here has
+; to be declared before it is defined.
+;
+; Every case below is the one the model-smt stage generates today, copied as it
+; stands, so that what is compiled from this file is what is compiled now. Each
+; program declares the parameters its own cases name, and no others.
+;
+; What stays in $eo_to_smt_type itself is the 6 cases that belong to no symbol
+; of the input: the function type, the ones a datatype or an uninterpreted sort
+; is built by ($eot_DatatypeType, $eot_DatatypeTypeRef, $eot_DtcAppType,
+; $eot_USort), and the default case.
+;
+; What stays in $eo_to_smt itself is the 11 cases that belong to no symbol: the
+; constructors of a Eunoia literal ($eot_bool, $eot_numeral, $eot_rational,
+; $eot_string, $eot_binary), those of a term the input declares ($eot_Var,
+; $eot_UConst, $eot_DtCons, $eot_DtSel), the application of one term to
+; another, and the default case, which gives $sm_none to everything else.
+;
+; This file is not included by anything yet.
+
+; -----------------------------------------------------------------------------
+; What this file contributes to, and the helpers its cases name
+; -----------------------------------------------------------------------------
+
+; The two aggregates this file contributes to, $eo_to_smt and
+; $eo_to_smt_type, are declared by plugins/model_smt/model_smt.eo, which is
+; what lets a case below recurse into them.
+
+; A helper a case names belongs to the symbol that names it and is written
+; out just above that symbol below, the first time it is needed.
+
+; -- ite
+(program $eoc_transform_ite
+  ((T Type) (T1 Type) (T2 Type) (T3 Type) (x1 T1) (x2 T2) (x3 T3))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_ite (ite x1 x2 x3)) ($sm_ite ($eo_to_smt x1) ($eo_to_smt x2) ($eo_to_smt x3)))
+  )
+)
+
+; -- not
+(program $eoc_transform_not
+  ((T Type) (T1 Type) (x1 T1))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_not (not x1)) ($sm_not ($eo_to_smt x1)))
+  )
+)
+
+; -- or
+(program $eoc_transform_or
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_or (or x1 x2)) ($sm_or ($eo_to_smt x1) ($eo_to_smt x2)))
+  )
+)
+
+; -- and
+(program $eoc_transform_and
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_and (and x1 x2)) ($sm_and ($eo_to_smt x1) ($eo_to_smt x2)))
+  )
+)
+
+; -- =>
+(program $eoc_transform_=>
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_=> (=> x1 x2)) ($sm_=> ($eo_to_smt x1) ($eo_to_smt x2)))
+  )
+)
+
+; -- xor
+(program $eoc_transform_xor
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_xor (xor x1 x2)) ($sm_xor ($eo_to_smt x1) ($eo_to_smt x2)))
+  )
+)
+
+; -- =
+(program $eoc_transform_=
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_= (= x1 x2)) ($sm_= ($eo_to_smt x1) ($eo_to_smt x2)))
+  )
+)
+
+; -- distinct
+(program $eo_to_smt_distinct_pairs
+  ((T Type) (s $smt_Term) (x T) (xs (@@TypedList T)))
+  :signature ($smt_Term (@@TypedList T)) $smt_Term
+  (
+  (($eo_to_smt_distinct_pairs s (@@TypedList.cons x xs))
+     ($sm_and ($sm_not ($sm_= s ($eo_to_smt x))) ($eo_to_smt_distinct_pairs s xs)))
+  (($eo_to_smt_distinct_pairs s (@@TypedList.nil T)) $sm_true)
+  (($eo_to_smt_distinct_pairs s xs) $sm_none)
+  )
+)
+(program $eo_to_smt_distinct
+  ((T Type) (s $smt_Term) (x T) (xs (@@TypedList T)))
+  :signature ((@@TypedList T)) $smt_Term
+  (
+  (($eo_to_smt_distinct (@@TypedList.cons x xs))
+     ($sm_and ($eo_to_smt_distinct_pairs ($eo_to_smt x) xs) ($eo_to_smt_distinct xs)))
+  (($eo_to_smt_distinct (@@TypedList.nil T)) $sm_true)
+  (($eo_to_smt_distinct xs) $sm_none)
+  )
+)
+(program $eo_to_smt_typed_list_elem_type ((T Type) (t T) (ts (@@TypedList T)))
+  :signature (T) $smt_Type
+  (
+  (($eo_to_smt_typed_list_elem_type (@@TypedList.nil T))
+    (eo::define ((eT ($eo_to_smt_type T)))
+      ($native_ite ($smtx_type_wf eT) eT $tsm_none)))
+  (($eo_to_smt_typed_list_elem_type (@@TypedList.cons t ts))
+    (eo::define ((eT ($smtx_typeof ($eo_to_smt t))))
+      ($native_ite ($native_Teq eT ($eo_to_smt_typed_list_elem_type ts))
+        eT
+        $tsm_none)))    
+  (($eo_to_smt_typed_list_elem_type t) $tsm_none)
+  )
+)
+(program $eoc_transform_distinct
+  ((T Type) (T1 Type) (x1 T1))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_distinct (distinct x1)) ($native_ite ($native_Teq ($eo_to_smt_typed_list_elem_type x1) $tsm_none) $sm_none ($eo_to_smt_distinct x1)))
+  )
+)
+
+; -- @purify
+(program $eoc_transform_@purify
+  ((T Type) (T1 Type) (x1 T1))
+  :signature (T) $smt_Term
+  (
+  (($eoc_transform_@purify (@purify x1)) ($sm_@purify ($eo_to_smt x1)))
+  )
+)
+
+; -- +
+(program $eoc_transform_+
+  ((T Type) (T1 Type) (T2 Type) (x1 T1) (x2 T2))
+  :signature (T) $smt_Term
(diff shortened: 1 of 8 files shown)


plugins/model_smt/model_smt.h
```diff
@@ -0,0 +1,126 @@
...
+class ModelSmt : public StdPlugin
+{
+ public:
+  /**
+   * Construct the plugin. defsFile names the signature of the *input* written
+   * in the deep embedding, which says what each of its symbols means to the
+   * model; the SMT-LIB signature it is written against is fixed, being the
+   * target rather than a matter of the input. See loadDefs.
+   */
+  ModelSmt(State& s, const std::string& defsFile);
```
daniel-larraz, line 72: **Blocking: `ModelSmt` cannot be instantiated by the plugin factory.**

`createPlugin` does `make_unique<ETHOS_PLUGIN_CLASS>(state)` — one argument (`src/plugin.cpp:25`), and that is the only construction path `add_ethos_plugin_executable` wires up. `ModelSmt(State&, const std::string&)` has no default for `defsFile`, so compiling `src/plugin.cpp` with `ETHOS_PLUGIN_CLASS=ModelSmt` fails:

```
error: no matching function for call to 'ethos::ModelSmt::ModelSmt(ethos::State&)'
note: candidate expects 2 arguments, 1 provided
```

`plugins-compile-check` does not catch this: it builds the plugin objects but never `src/plugin.cpp`.

Relatedly, no `--defs` option exists anywhere in `src/`, though the docs reference it in three places, so `d_defsFile` currently has no way to become non-empty.

`Desugar(State& s, bool genVcs = false)` is the precedent — giving `defsFile` a default would make the plugin selectable today.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+      for (const std::string& u : b->d_uses)
+      {
+        std::map<std::string, const DefsBlock*>::const_iterator it =
+            owner.find(u);
+        if (it != owner.end() && declared.count(it->second->d_sym) == 0
+            && placed.insert(it->second).second)
+        {
+          byDecl.push_back(it->second);
+        }
+      }
```
daniel-larraz: **Blocking: dependency pre-placement is only one level deep — reachable use-before-define.**

This loop walks `d_uses` of a *declared* symbol's block and places undeclared owners before it, but never recurses into those owners' own uses. A block reached only as a dependency-of-a-dependency is therefore missed here and falls through to the trailing `for (blocks) if (placed.insert(b).second)` below, which appends it at the **very end** of `byDecl` — after every case and program that needs it. `$smtx_typeof_*` and `$eo_to_smt*` aux programs have no forward declaration, so that is a hard ordering error in the generated file.

I reproduced `loadDefs` verbatim in a standalone harness and ran it with subsets of the declared symbols. Declaring every 3rd `cpc_defs.eo` symbol produces, in the generated output:

```
$smtx_elem_typeof_seq_value: first use line 1535, defined line 1546
```

Making the placement recursive gives 0 violations at every subset I tried (full set, every 2nd, every 3rd, every 4th):

```cpp
std::function<void(const DefsBlock*)> place = [&](const DefsBlock* x) {
  for (const std::string& u : x->d_uses) {
    auto it = owner.find(u);
    if (it != owner.end() && declared.count(it->second->d_sym) == 0
        && placed.insert(it->second).second) {
      place(it->second);
      byDecl.push_back(it->second);
    }
  }
};
place(b);
```

The current CPC input happens not to trigger this, but the whole point of `--defs` being a parameter is that the input signature varies.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+  replace(finalSmt, "$EO_DESUGAR_AUX$", d_desugarAux.str());
+  replace(finalSmt, "$EO_TO_SMT_CASES$", d_eoToSmt.str());
+  replace(finalSmt, "$EO_TO_SMT_TYPE_CASES$", d_eoToSmtType.str());
+  replace(finalSmt, "$SMT_TERM_CONSTRUCTORS$", d_smtTerms.str());
+  replace(finalSmt, "$SMT_TYPEOF_CASES$", d_smtTypeof.str());
+  replace(finalSmt, "$SMT_TYPEOF_AUX$", d_smtTypeofAux.str());
+
+  std::string outPath = getOutputPath("plugins/model_smt/model_smt_gen.eo");
+  std::ofstream oute(outPath);
+  oute << finalSmt;
```
daniel-larraz: **Blocking: neither the template read nor the output write is checked.**

If `plugins/model_smt/model_smt.eo` cannot be opened, `finalSmt` is empty, all ten `replace` calls silently no-op, and an empty `model_smt_gen.eo` is written with no diagnostic. If `outPath` cannot be opened, nothing is written and nothing is said.

Every other place in the tree guards this — `StdPlugin::copyResourceToOutput`, `MetaReducePlugin` (`plugins/meta_reduce_plugin.cpp:312,342,348`), `DesugarChecker` (`plugins/desugar/desugar_checker.cpp:144`) all `EO_FATAL`. Given that `finalizeDecl` aborts on an unmodeled symbol *for soundness*, silently emitting an empty model-semantics file is the wrong failure mode here.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+  auto replace = [](std::string& txt,
+                    const std::string& tag,
+                    const std::string& replacement) {
+    const std::string guarded = ";" + tag;
+    auto pos = txt.find(guarded);
+    if (pos != std::string::npos)
+    {
+      txt.replace(pos, guarded.length(), replacement);
+    }
+  };
```
daniel-larraz: **Blocking: a missing placeholder silently discards a whole stream.**

If a `;$TAG$` line is typo'd or dropped from the template, `find` returns `npos`, the branch is skipped, and the run still succeeds — having lost, say, every `$smtx_typeof` case. This should be `EO_FATAL`, for the same soundness reason `finalizeDecl` is.

This also duplicates the existing `StdPlugin::replace`; `StdPlugin::replace(txt, ";" + tag, r)` would do, modulo the missing-tag check.

While here: a cheap catch-all for this class of bug is to assert after all ten substitutions that no `$eoc_` name survived into `finalSmt` — see the `casesOf` comment.

plugins/model_smt/defs_reader.cpp
```diff
@@ -0,0 +1,402 @@
...
+      while (pos != std::string::npos)
+      {
+        out = out.substr(0, pos) + to + out.substr(pos + from.size());
+        pos = out.find(from, pos + to.size());
+      }
+      b.d_desugarAux.push_back(out);
+    }
+    else if (isPre("$eoc_eval_"))
+    {
+      b.d_evalCases = casesOf(f, name, "$smtx_model_eval");
```
daniel-larraz: Per-symbol case lists are **assigned**, not appended — same for `d_typeofCases`, `d_transCases` and `d_transTypeCases` just below. A block containing two `$eoc_eval_*` programs would silently drop the first.

No block has two today (I checked all four prefixes across both `.eo` files), but the silence is the problem: either append, or assert the list is empty first.

plugins/model_smt/defs_reader.cpp
```diff
@@ -0,0 +1,402 @@
...
+  std::vector<std::string> ret;
+  if (body.empty())
+  {
+    return ret;
+  }
+  const std::string& cl = body.back();
+  for (const std::string& c : forms(cl.substr(1, cl.size() - 2)))
+  {
+    std::string cs = c;
+    size_t p = cs.find(name);
```
daniel-larraz, line 188: `cs.find(name)` rewrites only the **first** substring occurrence of the program name. A self-recursive `$eoc_eval_X` — where the case body calls `$eoc_eval_X` again — would leave a reference to a program that is never emitted, silently, in the generated file.

Currently clean (0 `$eoc_` names survive substitution for the CPC input, which I checked on the generated output), but nothing enforces it. A one-line check in `ModelSmt::finalize` that `finalSmt.find("$eoc_") == std::string::npos` would make this and the missing-placeholder case loud.

plugins/model_smt/defs_reader.cpp
```diff
@@ -0,0 +1,402 @@
...
+    b.d_uses.erase(d);
+  }
+  for (const std::string& d : b.d_defs)
+  {
+    d_owner[d] = d_blocks.size();
+  }
+  d_blocks.push_back(b);
+}
+
+bool DefsFile::read(const std::string& path)
```
daniel-larraz, line 323: `read` returns `true` for a file that yields zero blocks — a file that exists but has no `; -- ` markers, e.g. the wrong file passed to `--defs`, is indistinguishable from success here. The failure then surfaces much later as a confusing `no model semantics found for <first symbol>` abort.

Returning false (or `EO_FATAL`-ing) when `d_blocks` is empty would point at the actual problem.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+    EO_FATAL() << "ModelSmt: could not read the SMT-LIB signature written in"
+                  " the deep embedding";
+  }
+  if (d_defsFile.empty())
+  {
+    EO_FATAL() << "ModelSmt: no signature was given for the input; name the"
+                  " one written in the deep embedding with --defs, e.g."
+                  " plugins/model_smt/cpc_defs.eo for CPC";
+  }
+  if (!d_inputDefs.read(d_defsFile))
```
daniel-larraz, line 80: Asymmetric path handling: `smt_defs.eo` goes through `getResourcePath` (line 62) but `d_defsFile` is used raw. The class doc suggests `plugins/model_smt/cpc_defs.eo` as a `--defs` value, which only resolves if cwd happens to be the repo root. Worth either routing it through `getResourcePath` too, or documenting that `--defs` takes an absolute/cwd-relative path.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+  // a proof may write, so no model has to say anything about it.
+  if (name.compare(0, 1, "$") == 0 || name.compare(0, 2, "@@") == 0)
+  {
+    return;
+  }
+  // This assertion is critical for soundness: if we do not know how to
+  // interpret the symbol, we cannot claim this verification condition
+  // accurately models SMT-LIB semantics.
+  EO_FATAL() << "ERROR: no model semantics found for " << name;
+  Assert(false) << "No model semantics found for " << name;
```
daniel-larraz: `Assert(false)` here is dead code: `~FatalStream` is `[[noreturn]]` and aborts, so `EO_FATAL()` on the previous line never returns. (And under a non-`EO_ASSERTIONS` build `Assert` compiles to a no-op anyway.) The `EO_FATAL` alone carries the intent.

plugins/model_smt/defs_reader.h
```diff
@@ -0,0 +1,89 @@
...
+  std::vector<const DefsBlock*> select(const std::set<std::string>& syms,
+                                      const std::set<std::string>& names =
+                                          {}) const;
+  /** The names the blocks use that no block of this file defines. */
+  std::set<std::string> externalUses(
+      const std::vector<const DefsBlock*>& blocks) const;
+  /** The blocks, in the order the file gives them. */
+  const std::vector<DefsBlock>& getBlocks() const { return d_blocks; }
+  /** True if some block is of the symbol sym. */
+  bool hasSymbol(const std::string& sym) const;
```
daniel-larraz: `hasSymbol` is declared and defined (`defs_reader.cpp:330`) but never called anywhere in the tree.

plugins/model_smt/defs_reader.h
```diff
@@ -0,0 +1,89 @@
+#ifndef PLUGINS__MODEL_SMT__DEFS_READER_H
```
daniel-larraz, line 10: `defs_reader.h` and `defs_reader.cpp` are the only two files in this PR without the project license header that every other file carries.

plugins/model_smt/model_smt.cpp
```diff
@@ -0,0 +1,256 @@
...
+void ModelSmt::bind(const std::string& name, const Expr& e)
+{
+  if (e.getKind() != Kind::CONST)
+  {
+    return;
+  }
+  d_declSeen.emplace_back(name, e);
+}
+
+void ModelSmt::finalizeDecl(const std::string& name, const Expr& e)
```
daniel-larraz: `const Expr& e` is unused in the body. Either drop it or `(void)e`.

plugins/model_smt/defs_reader.cpp
```diff
@@ -0,0 +1,402 @@
...
+    }
+    if (form[i] == ';')
+    {
+      while (i < n && form[i] != '\n')
+      {
+        i++;
+      }
+      continue;
+    }
+    if (form[i] == '$')
```
daniel-larraz, line 157: `namesOf` treats every `$`-word as a used *name*, which includes bound program parameters — `$T`, `$i`, `$wm1`, `$wm2` all show up in `d_uses` for the current files.

Conservative for `select` (it over-selects), but it also feeds `externalUses` and the `owner` lookup in `ModelSmt::loadDefs`, so it would misfire if any block ever defined a name matching a common parameter. Worth at least a comment noting the over-approximation is deliberate.

plugins/model_smt/defs_reader.cpp
```diff
@@ -0,0 +1,402 @@
...
+  std::ifstream in(path);
+  if (!in.is_open())
+  {
+    return false;
+  }
+  std::stringstream ss;
+  ss << in.rdbuf();
+  const std::string text = ss.str();
+  // A block runs from the line that names its symbol to the next one.
+  const std::string mark = "\n; -- ";
```
daniel-larraz, line 341: The marker is `"\n; -- "`, so a block starting on line 1 of the file is skipped. Theoretical today — both `.eo` files open with prose — but a leading `\n` requirement is easy to trip over when these files start being generated.

plugins/model_smt/smt_defs.eo
```diff
@@ -0,0 +1,3818 @@
+; The SMT-LIB signature written *directly* in the deep embedding.
+;
+; This is what the model-smt stage compiles the SMT-LIB half of a model from.
+; For each of the 128 symbols it covers it gives:
```
daniel-larraz: Stale count: the file has 149 `; -- ` blocks, 130 of which are non-helper (non-`$`) symbols. Neither is 128.

plugins/model_smt/cpc_defs.eo
```diff
@@ -0,0 +1,2091 @@
+; The CPC signature written *directly* as a transformation into the deep
+; embedding.
+;
+; This is what the model-smt stage compiles the CPC half of a model from. For
+; each of the 180 symbols it covers it gives one thing:
```
daniel-larraz: Stale count: the file has 187 `; -- ` blocks, 181 of which are non-helper (non-`$`) symbols. Neither is 180.

daniel-larraz (commented):
## Review

Reviewed the C++ (`defs_reader.*`, `model_smt.*`, ~870 lines) closely; treated the three `.eo` files (~7400 lines) as data and exercised them through a standalone harness rather than auditing their semantics line by line.

### What I verified

- `make plugins-compile-check` builds clean. The CMake `GLOB_RECURSE` picks the new sources up automatically, so no CMake change is needed here.
- All 10 `;$...$` placeholders appear in `model_smt.eo` exactly once, and `$SMT_TYPEOF_AUX$` / `$EO_TO_SMT_AUX$` correctly precede their case placeholders.
- Block parsing loses nothing: 149/149 and 187/187 `; -- ` markers become blocks, no duplicate symbols in either file.
- Generating with the full CPC declaration set produces no dangling `$eoc_*` references and no use-before-define among the 681 names defined in the output. Every unresolved `$`-name is either a bound program parameter or something supplied by the template / desugar output / native embedding.

### Findings

4 blocking, 4 worth fixing, 7 minor — filed inline. The two that matter most:

1. **`ModelSmt` cannot be constructed by `createPlugin`** — the 2-argument constructor does not match the factory, and it is a compile error rather than a runtime one. `plugins-compile-check` misses it because it never builds `src/plugin.cpp`.
2. **Transitive dependency ordering in `loadDefs` is one level deep**, which puts a dependency-of-a-dependency at the end of the emission order. I reproduced a concrete use-before-define (`$smtx_elem_typeof_seq_value`) with a subset of declared symbols, and verified that making the placement recursive fixes it at every subset I tried.

The other blocking two are both silent-failure modes — unchecked file I/O in `finalize()`, and `replace` dropping a whole stream when a placeholder is absent. Both sit oddly against `finalizeDecl`'s explicit soundness abort.

### Process

- Merge base is `216364d6`; `main` is at `ddd9184d`. Worth rebasing.
- No tests. `DefsFile` is self-contained, pure text-in/structure-out, and links against nothing but `base/check.h` — it is cheap to unit-test, and finding 2 is exactly the kind of bug a test over a couple of declaration subsets would have caught.

ajreynol:
I addressed the review, note that I am not yet adding local tests for this plugin as its full functionality is not yet available on main.

daniel-larraz:
All 16 threads addressed — resolved them. I re-ran the two blocking ones rather than reading the diff:

- **Factory constructor**: compiling `src/plugin.cpp` with `ETHOS_PLUGIN_CLASS=ModelSmt` now succeeds where it previously errored, and the `static_assert` means it can't regress silently.
- **Transitive ordering**: 0 use-before-define at every k-th declaration subset (k=1..5), where k=3 previously put `$smtx_elem_typeof_seq_value` 11 lines after its first use. A 30-trial randomized subset sweep gives 0/30 violations on the new code vs 3/30 on the old, so the check is sensitive rather than vacuous.

Also confirmed `make plugins-compile-check` is clean, that full-input generation is byte-identical to before (so the `casesOf` replace-all and the `read` leading-newline change are behavior-preserving on real data), and that the new `$eoc_` guard can't false-positive — the template contains zero `$eoc_` occurrences.

Two cosmetic leftovers, neither blocking:

1. `model_smt.h:36` still says the input signature is "which `--defs` names". There is still no `--defs` option, and the default constructor now hardcodes `cpc_defs.eo`, so the sentence points at something that doesn't exist yet. It also sits a little oddly against the claim two paragraphs up that the plugin "names no symbol of any theory". Worth rewording whenever the option actually lands.
2. `replacePlaceholder` searches for `guarded` to decide whether to fatal, then calls `StdPlugin::replace`, which searches for it again. Harmless, just a double `find`.

daniel-larraz (approved):

PR #228, opened 2026-08-24
Add lean meta plugin

ajreynol:
This defines the compilation from Eunoia to Lean. It is run after the model smt embedded semantics are written to the Eunoia.

This also includes some static configuration files specific to cpc, which may move ownership later.

### DIFF
--- a/plugins/lean_meta/cpc_termination.lean
+++ b/plugins/lean_meta/cpc_termination.lean
@@ -1,0 +1,46 @@
+-- Termination clauses for the programs of the CPC signature.
+--
+-- This is the file --lean-config names when the input is CPC; the format and
+-- what the clauses are for are described in plugins/lean_meta/termination.lean,
+-- which holds the ones for the deep embedding itself.
+
+-- Distinctness of a list of values is decided by walking the list and, at each
+-- element, the rest of it, so the measure is the size of the two lists.
+
+-- $set_is_not_subset $seq_distinct_terms
+termination_by x1 x2 x3 => sizeOf x1 + sizeOf x2
+
+-- $dt_distinct_terms
+termination_by x1 x2 => sizeOf x1 + sizeOf x2
+
+-- $are_distinct_terms_type
+termination_by x1 x2 x3 => sizeOf x1 + sizeOf x2 + 1
+decreasing_by
+  all_goals simp_wf
+  all_goals omega
+
+-- Flattening a regular expression alternates between the tree and a pass over
+-- it, which the flag distinguishes, so the measure counts the tree twice and
+-- breaks the tie with the flag.
+
+-- $re_flatten
+termination_by flag tree => 2 * sizeOf tree + (if flag = Term.Boolean true then 1 else 0)
+decreasing_by
+  all_goals simp_wf
+  all_goals omega
+
+-- Regular expression inclusion descends through four mutually recursive
+-- helpers, so each step is given room by scaling the measure, and the base
+-- case is ordered after the others by the offset.
+
+-- $str_re_includes_lhs_union $str_re_includes_lhs_star $str_re_includes_rhs_inter $str_re_includes_rec
+termination_by a b => 4 * (sizeOf a + sizeOf b)
+decreasing_by
+  all_goals simp_wf
+  all_goals omega
+
+-- $str_re_includes_base_rec
+termination_by a b => 4 * (sizeOf a + sizeOf b) + 1
+decreasing_by
+  all_goals simp_wf
+  all_goals omega
--- a/plugins/lean_meta/lean_meta_checker.lean
+++ b/plugins/lean_meta/lean_meta_checker.lean
@@ -28,14 +28,14 @@ def native_str_substr (s : native_String) (i n : native_Int) : native_String :=
     let start : Nat := Int.toNat i
     let take  : Nat := Int.toNat (min n (len - i))
     (s.drop start).take take
-def native_str_indexof_rec (s t : native_String) (i len fuel : Nat) : native_Int :=
+def native_str_indexof_rec (s t : native_String) (i fuel : Nat) : native_Int :=
   match fuel with
   | 0 => -1
   | fuel + 1 =>
       if native_string_prefix_eq t (s.drop i) then
         Int.ofNat i
       else
-        native_str_indexof_rec s t (i + 1) len fuel
+        native_str_indexof_rec s t (i + 1) fuel
 def native_str_indexof (s t : native_String) (i : native_Int) : native_Int :=
   if i < 0 then
     -1
@@ -44,7 +44,7 @@ def native_str_indexof (s t : native_String) (i : native_Int) : native_Int :=
     let start := Int.toNat i
     let tLen := Int.toNat (native_str_len t)
     if h : start + tLen <= sLen then
-      native_str_indexof_rec s t start tLen (sLen - (start + tLen) + 1)
+      native_str_indexof_rec s t start (sLen - (start + tLen) + 1)
     else
       -1
 
@@ -66,9 +66,11 @@ def native_tcmp (a b : Term) : native_Bool :=
   | Ordering.lt => true
   | _ => false
 
-/- Used for defining hash -/
+/- Used for defining hash. This is intentionally a stub: EO treats hash as an
+   underconstrained oracle, so signatures must not rely on distinct terms
+   receiving distinct values in the executable Lean checker. -/
 def native_thash : Term -> native_Int
-  | _ => 0 -- FIXME
+  | _ => 0
 
 /- Proofs -/
 inductive Proof : Type where
--- a/plugins/lean_meta/lean_meta_checker.lean
+++ b/plugins/lean_meta/lean_meta_checker.lean
@@ -1,0 +1,149 @@
+module
+
+public import $EO_CALC$.LogosTerm
+import all $EO_CALC$.LogosTerm
+public import $EO_CALC$.SmtEval
+import all $EO_CALC$.SmtEval
+
+public section
+
+set_option linter.unusedVariables false
+set_option maxHeartbeats 10000000
+
+namespace Eo
+
+open SmtEval
+
+/- Eunoia literal evaluation defined -/
+
+def native_str_len : native_String -> native_Int
+  | x => Int.ofNat x.length
+def native_str_concat : native_String -> native_String -> native_String
+  | x, y => x ++ y
+def native_str_substr (s : native_String) (i n : native_Int) : native_String :=
+  let len : Int := (native_str_len s)
+  if i < 0 || n <= 0 || i >= len then
+    []
+  else
+    let start : Nat := Int.toNat i
+    let take  : Nat := Int.toNat (min n (len - i))
+    (s.drop start).take take
+def native_str_indexof_rec (s t : native_String) (i len fuel : Nat) : native_Int :=
+  match fuel with
+  | 0 => -1
+  | fuel + 1 =>
+      if native_string_prefix_eq t (s.drop i) then
+        Int.ofNat i
+      else
+        native_str_indexof_rec s t (i + 1) len fuel
+def native_str_indexof (s t : native_String) (i : native_Int) : native_Int :=
+  if i < 0 then
+    -1
+  else
+    let sLen := Int.toNat (native_str_len s)
+    let start := Int.toNat i
+    let tLen := Int.toNat (native_str_len t)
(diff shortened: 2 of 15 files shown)


daniel-larraz (commented):
Reviewed at `9a05f39e` (main merged). `plugins-compile-check` builds clean on the merged head. I read the Lean templates against the SMT-LIB semantics that #227 brings in, since those two now have to agree.

Nice work overall — the separation of the termination clauses into their own file, and the comment explaining why they're Lean text rather than something the plugin derives, is a good call. Findings below, most severe first.

## Correctness

**1. The step-pop case calls a theorem the template never emits** — `plugins/lean_meta/lean_meta_reduce.cpp:1793-1826`

`printStepCase` with `isPop = true` builds `thmName` as `cmd_step_pop_<rule>_properties` and emits a call taking `A root args premises hATrans hATy hPremisesTrans hPremisesTy hResultTy`. But it writes that file from `plugins/lean_meta/lean_meta_rule.lean`, which declares only:

```lean
public theorem cmd_step_$EO_RULE$_properties
    (M : SmtModel) (hM : model_total_typed M)
    (s : CState) (args : CArgList) (premises : CIndexList) :
```

No `pop_` infix, and a different signature. `scope` is the one rule that takes the pop branch, so it generates a reference to a name that doesn't exist. Both branches also write the same `lean_meta_rule_<prule>_gen.lean` path, so the emitted theorem can't vary with `isPop`. Either the template needs a pop variant selected by `isPop`, or the call has to be built without the infix.

**2. `printStepEmptyCase` is entirely commented out** — `lean_meta_reduce.cpp:1836-1847`

The body is dead comments; `out` and `isPop` are unused. `d_rlIncludeStep` / `d_rlIncludeStepPop` therefore get no case for the complementary side of each rule, and the generated match is non-exhaustive.

**3. Leftover debug print duplicates every echo** — `lean_meta_reduce.cpp:1883`

```cpp
std::cout << "ECHO " << msg << std::endl;
```

sits *before* the `lean-meta` branch, so it fires for every echo the plugin doesn't consume. Those return `true`, and `State::echo` (`src/state.cpp:1795`) then prints `msg` a second time. There's no `MetaReducePlugin::echo` override, so `Plugin::echo`'s default `return true` applies.

This is live against the merged signatures — `plugins/model_smt/cpc_defs.eo:2087-2091` has five `(echo "eoc-exclude ...")` commands and `plugins/desugar/eo_desugar.eo:652` has `(echo "include model_smt :deps ...")`. Running lean_meta over them produces eleven lines of doubled output.

**4. String literals are interpolated into Lean unescaped** — `lean_meta_reduce.cpp:338` and `:385`

```cpp
os << "(native_string_lit \"" << css << "\")";
```

`css` comes from `Literal::toString()`, which renders backslashes and non-printables as `\u{5c}`-style codes and leaves `"` untouched (`src/util/string.cpp:197`). Lean 4 accepts neither: an embedded `"` terminates the literal, and `\u{...}` is not Lean escape syntax. `quoteLeanString` already exists and is used for parser op/rule names (`:1464`, `:1530`) — it just isn't applied on these two paths. It would also need a `\u{}` case to be complete.

## Dead code

**5. `isSimple` can never become true** — `lean_meta_reduce.cpp:792-841`

It's initialized `false` and the only two assignments in the block also set it `false`. Consequences, both live now:

- `d_simpleDefProgs` is never populated, so `$LEAN_EO_IS_OBJ_SIMPLE_DEFS$` always expands to nothing.
- `isSimple` is read again at `:864` to route `$eo_to_smt*`, and #227 makes `$eo_to_smt` / `$eo_to_smt_type` real (`plugins/model_smt/defs_reader.cpp:278-300`). Every such definition unconditionally lands in `d_eoIsObjDefs`.

Two further slips in the same block that will bite once the initializer is fixed:

- `:813` tests `e != prog`, but `calls` holds `Kind::PROGRAM_CONST` subterms while `prog` is the `Kind::PROGRAM` body. It should compare against `v`, the way `d_partialDefProgs.insert(v)` at `:850` does.
- `:839` inserts `prog` (the body) into `d_simpleDefProgs`, which `:813` then queries with a `PROGRAM_CONST`. Never matches.

**6. The `$eoo_` overload revert is a no-op** — `lean_meta_reduce.cpp:1226-1232`

It rewrites `cnamek`, but `cname` was derived from it at `:1156` and already printed at `:1224`, and `cnamek` isn't read again in the function. Overloaded constructors keep their mangled `__eoo_...` names.

## Latent / fragility

**7. `d_embedTOpDt[uarity]` can write out of bounds** — `lean_meta_reduce.cpp:1213-1215`

The array is `std::stringstream d_embedTOpDt[4]` (`lean_meta_reduce.h:207`) and `uarity` is `getNumChildren() - 1` of an opaque operator's type, which is unbounded. The guard is `Assert(uarity < 4)`, which compiles out in release — an operator with 4+ index arguments corrupts memory rather than failing. `AlwaysAssert` at minimum.

**8. Unguarded `substr` in the `lean-meta` echo handler, and an off-by-one** — `lean_meta_reduce.cpp:1903-1908`

`progName.substr(10)` throws `std::out_of_range` on a short name, and `fileName[0] = toupper(...)` writes into a possibly-empty string.

The offset also looks wrong. `:858` establishes the `$eo_model` convention, and `cleanSmtId` maps `$` → `__`, so `$eo_model_scope` cleans to `__eo_model_scope` — the prefix is 11 chars. `substr(10)` leaves `_scope`, so `prule == "scope"` never matches and the import becomes `Rules._scope`. Note this whole path is unexercised today: nothing in tree emits a `lean-meta` echo yet, though #226 documented the format at `plugins/trim_defs/trim_defs.cpp:498`. It's also the path finding 1 flows through.

**9. `substr(8)` for the 6-char `$eo.l.` prefix** — `lean_meta_reduce.cpp:819-822`

Linear-pattern programs are named `$eo.l.<n>.<prog>` (`plugins/linear_patterns/linear_patterns.cpp:54`). `substr(8)` assumes a single-digit `n`; at `n >= 10` `rawName` becomes `0.$prog`. Only feeds the `d_partialExc` lookup, which is empty by default, so it's inert today. Finding the 4th `.` would be robust.

**10. Null deref after a compiled-out assert** — `lean_meta_reduce.cpp:363-370`

`getEmbedName` does `Assert(false)` when `isSmtApplyApp` fails, then unconditionally dereferences `oApp[1].getValue()->asLiteral()`. In release that's a null deref rather than a diagnostic.

## Minor

- `native_thash` returns `0` for every term (`lean_meta_checker.lean:75`, marked `FIXME`) — every `eo::hash` collides. Fine as a stub, but worth a note that any signature relying on hash distinctness will silently misbehave.
- `native_binary_extract` (`lean_meta_smt_eval.lean:165`) ignores its `w` and `x1` parameters. That's correct — `smt_defs.eo:845` masks the result to `2^(x1-x2+1)` at the call site — but a one-line comment saying the caller owns the width would save the next reader the trip.
- `native_str_indexof_rec`'s `len` parameter is unused (`lean_meta_checker.lean:26`).

## What I checked and found clean

- **The `native_binary_*` / `native_pi*` encoding.** I initially read the `.toInt` in `native_piand`/`pior`/`pixor` (`lean_meta_smt_eval.lean:92-97`) as a signedness bug, but it isn't: `smt_defs.eo` wraps every use in `mod_total ... (int.pow2 w)`, and `native_mod_total` is Lean's `%` on `Int` (`Int.emod`, always in `[0, |n|)` for a positive modulus), so the signed intermediate is normalized back to the unsigned residue. `w = 0` is guarded by the `native_ite (native_zeq w 0) 0 ...` wrapper. Same story for `binary_not`, `binary_concat`, `binary_uts`, `binary_udiv` — the helpers are deliberately partial in range and the signature owns the masking. `binary_uts` also checks out as the correct unsigned-to-signed reinterpretation.
- **Native op coverage.** I extracted all 129 `$native_apply_N "op"` names used by `model_smt/*.eo` and `desugar/*.eo`, applied `cleanSmtId`'s rewrite table, and confirmed every one resolves to a `native_*` definition in the Lean templates — including `const_id`, which #226 added after this branch was cut (`plugins/desugar/native_embed.eo:146` ↔ `lean_meta_smt_model.lean:112`). No gaps.
- **Template tags.** Every `$TAG$` the C++ substitutes has a matching placeholder in its template, and vice versa.
- **String builtins.** `native_str_substr` and `native_str_indexof` match SMT-LIB on the edge cases I traced — negative index, `n <= 0`, empty needle at `|s|`.

I did not run Lean: the templates reference an `$EO_CALC$` package that isn't in this repo, so findings 1, 2 and 4 are read from the source rather than from a failing elaboration.

---
🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

daniel-larraz (commented):
Re-reviewed at `bc0b7924`. `plugins-compile-check` still builds clean. Thanks — 8 of the 10 findings and all 3 minors are addressed. Two things still open, and one correction of my own to make.

## Correction to my previous review

My off-by-one claim about `progName.substr(10)` was wrong. I inferred an `$eo_model_` prefix from the check at `lean_meta_reduce.cpp:858`, but the actual prefix is `$eo_prog_`, and `cleanId("$eo_prog_")` is exactly 10 characters — the original `substr(10)` was already correct. The missing bounds guard was the real defect, and the new explicit `progPrefix` plus `EO_FATAL` is a better fix than what I described. Sorry for the noise.

## Still open

**1. String literal escaping** — `lean_meta_reduce.cpp:338` and `:385` are unchanged.

Both still emit the literal raw into Lean:

```cpp
os << "(native_string_lit \"" << css << "\")";
```

`css` comes from `Literal::toString()`, which leaves `"` untouched and renders backslashes and non-printables as `\u{5c}`-style codes (`src/util/string.cpp:197`). Lean 4 accepts neither — an embedded quote terminates the literal, and `\u{...}` is not Lean escape syntax. `quoteLeanString` is already there and already used for parser op/rule names at `:1464` and `:1530`; it needs a `\u{}` case added and then applying at these two sites. Any signature or proof carrying such a string currently produces Lean that doesn't parse.

**2. The non-exhaustive `cases r with` is still present.**

Deleting `printStepEmptyCase` removed the dead function, but it emitted nothing to begin with — the generated output is byte-identical before and after. The underlying issue is downstream, in `lean_meta_rule_lemmas.lean`:

- `:40` — `cases r with` / `$EO_RULE_LEMMA_STEP_CASES$`
- `:79` — `cases r with` / `$EO_RULE_LEMMA_STEP_POP_CASES$`

Neither has a catch-all, and `echo()` gives each rule *either* a step case *or* a step-pop case:

```cpp
if (prule == "scope") { printStepCase(d_rlIncludeStepPop, prule, true); }
else                  { printStepCase(d_rlIncludeStep,    prule, false); }
```

So `$EO_RULE_LEMMA_STEP_CASES$` has no `scope` case, and `$EO_RULE_LEMMA_STEP_POP_CASES$` has a case only for `scope`. Lean's `cases ... with` requires an alternative for every constructor, so as generated neither lemma will elaborate.

The old commented-out body of `printStepEmptyCase` looks like it had the intended shape:

```lean
| scope   => exact False.elim (hProg (by simp [__eo_cmd_step_proven]))
| contra  => exact False.elim (hProg (by simp [__eo_cmd_step_pop_proven]))
```

If this is deliberately deferred that's fine — but the deleted stub was the only marker for it, so it'd be worth a `TODO` at one of the two `cases r with` sites so it isn't lost.

## Confirmed fixed

- **Step-pop theorem** — the new `lean_meta_rule_pop.lean` lines up with the call site: 4 explicit arguments (`A root args premises`) and 5 hypothesis arrows matching `hATrans hATy hPremisesTrans hPremisesTy hResultTy`.
- **`isSimple`** — all three slips corrected together, and the semantics hold up. `lean_meta_spec.lean:25-29` places `$LEAN_EO_IS_OBJ_SIMPLE_DEFS$` before the `mutual` block, so "simple" means "definable outside it", and a single self-recursive definition is legitimately fine there — which is exactly what `e != v` now permits. Emission order is safe too, since a simple program only references programs already written into the same stream. The updated header comment matches the new behaviour.
- **`ECHO` line**, **`$eoo_` revert** (with the comment recording that mangling is deliberate for unary vs. binary `-`), **`AlwaysAssert(uarity < 4)`**, **`$eo.l.` separator lookup** (correct for any counter value now), **`getEmbedName` preconditions** — all good.
- **Minors** — `native_thash`, `native_binary_extract`, and the unused `len` parameter all handled.

---
🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

daniel-larraz (approved):

PR #229, opened 2026-08-25
Add core Eunoia compiler infrastructure

ajreynol:
This introduces the Eunoia compiler "eoc" which orchestrates the combination of 4 plugins to compile Eunoia to various targets (including Lean).

Further targets for smt2, sygus, desugared eo are forthcoming.

Followup will add further installation, documentation, testing.

Co-Authored-By: GPT 5.6 <noreply@openai.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 build
+build-eoc
+tools/eoc/out
+tools/eoc/__pycache__
 .vscode
 .DS_Store
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,12 +71,9 @@ add_subdirectory(tests)
 # which additionally links them; see plugins/CMakeLists.txt.
 #
 # cpp_compiler is skipped: it is a snapshot from an earlier version of ethos
-# and is known not to compile (see plugins/cpp_compiler/README.md). scratch
-# directories are skipped as well: they hold code fragments kept for reference,
-# which are not standalone translation units.
+# and is known not to compile (see plugins/cpp_compiler/README.md).
 file(GLOB_RECURSE plugins_SRC CONFIGURE_DEPENDS "plugins/*.cpp")
 list(FILTER plugins_SRC EXCLUDE REGEX "/plugins/cpp_compiler/")
-list(FILTER plugins_SRC EXCLUDE REGEX "/scratch/")
 if(plugins_SRC)
   add_library(plugins-compile-check OBJECT EXCLUDE_FROM_ALL ${plugins_SRC})
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,16 +60,23 @@ add_subdirectory(src)
 add_subdirectory(tests)
 
 # Compile-only check for the experimental extensions in plugins/. Those are
-# not part of ethos and produce nothing here; this target exists so that CI
-# can catch them bitrotting against src/. It is EXCLUDE_FROM_ALL, so an
-# ordinary build never sees it. Build it explicitly with:
+# not part of ethos and produce nothing here; this target exists so that they
+# can be checked for bitrot against src/ from an ordinary build tree. It is
+# EXCLUDE_FROM_ALL, so an ordinary build never sees it. Build it explicitly
+# with:
 #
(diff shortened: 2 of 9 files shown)


daniel-larraz:
Review of the branch. Findings below, most significant first. Line references are against `eoc-main`.

**What checks out:** the CMake refactor — `ethos_configure_target`'s switch from `CMAKE_CURRENT_LIST_DIR` to `CMAKE_CURRENT_SOURCE_DIR` is the right change for a function hoisted into a module (both resolve to `plugins/cpp_compiler` for the existing caller), `ethos_add_core_library` reproduces the previous glob/filter exactly, and `ETHOS_PLUGIN_OUTPUT_DIR` matches the driver's expected `<build-dir>/out/plugins/`. The `LEAN_OUTPUTS` table lines up with every `$EO_CALC$.<Module>` import in the templates except the per-rule one (finding 2).

---

### 1. `defs_blocks` silently drops a block that starts on line 1 — `tools/eoc/driver.py:292`

```python
for block in re.split(r"\n; -- ", self.defs_file.read_text())[1:]:
```

`DefsFile::read` (`plugins/model_smt/defs_reader.cpp:338`) prepends a newline before splitting on the same marker, with the comment *"Prepending a newline lets the same marker recognize a block on line one."* The Python side has no such trick, so for a `--defs` file whose first line is `; -- Sym`, that block's `eoc-exclude` and `depends` directives never reach the desugar/trim-defs stages — while model-smt still honors them, giving the two halves of the compiler different views of the same file. Latent today only because `cpc_defs.eo` happens to open with a comment header.

### 2. Per-rule Lean modules are published one directory level off — `tools/eoc/driver.py:242`

`publish_generated_lean_rule_outputs` writes to `<out>/lean/Rules/`, but `lean_meta_reduce.cpp:1908` emits `import $EO_CALC$.Proofs.Rules.<Name>`, which resolves to `<out>/lean/Proofs/Rules/`. Every other published module's path matches its import exactly, so this one is inconsistent. If the flat output tree is meant to be copied into an existing package layout rather than built as-is, that's worth a comment — but the asymmetry with the other outputs stands either way.

### 3. `--require-proof-of-false` and `--stats` are accepted but inert — `plugins/main_eoc.cpp:219`

Both are swallowed by the generic `opts.setOption` fallthrough in the argument loop, but the corresponding checks and `stats.toString(...)` output from `src/main.cpp:229` were not carried over. `main_eoc` ends with an unconditional `exit(0)`, so a proof that does not end in `false` passes silently and `--stats` prints nothing. Either implement them or reject them explicitly.

### 4. `--plugin.model-smt` without `--defs` bypasses the default defs file — `plugins/main_eoc.cpp:64`

`createPlugin` always calls the two-argument `ModelSmt(s, defsFile)` ctor. With no `--defs`, `defsFile` is empty, so the one-argument ctor's default `getResourcePath("plugins/model_smt/cpc_defs.eo")` is skipped and the run aborts late inside `loadDefs`, rather than either defaulting correctly or failing next to the other option validations at the end of the parse loop.

### 5. `lean --all INPUT rule1` silently ignores the positional targets — `tools/eoc/driver.py:487`

The `if all_targets:` branch never reads `targets`; the whole signature is compiled with no warning that the named rules were ignored.

### 6. The README's `lean` example exits 2 — `tools/eoc/README.md:37`

The documented `lean --defs ... Cpc.eo` invocation has neither `--all` nor a positional target, so `parser.error` fires.

### 7. Inconsistent path handling for file options — `tools/eoc/driver.py:643`

`--defs`, `--lean-config` and `--build-dir` use bare `Path(...)` while `input` goes through `resolve_path_arg`, so `~` is not expanded for the former. Combined with `defs_blocks`' `if ... not self.defs_file.exists(): return []`, a mistyped or unexpanded `--defs` path degrades to silently-empty excludes/depends instead of an error.

### 8. Stale docstring reference — `tools/eoc/driver.py:84`

Points at `tools/eoc/cpc/common.sh`, which is not in the tree.

---

1 and 3 look like the blocking ones; 2 and 4 are close behind. The rest are cleanups.

daniel-larraz (approved):

PR #230, opened 2026-08-26
Add smt meta plugin

ajreynol:
This gives compilation to smt2 and sygus. 

This PR also updates the CI to check more tests with plugins.

Note this functionality is not on the critical path.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,9 @@ jobs:
 
     # The experimental extensions in plugins/ are not part of the ethos build
     # and nothing else compiles them, so without this they would silently rot
-    # against src/. A debug build ensures assertions are compiled too.
+    # against src/. A debug build ensures assertions are compiled too. Building
+    # ethos-eoc covers every plugin source outside cpp_compiler/, so it
+    # subsumes the plugins-compile-check target of the top-level build.
     concurrency:
       group: plugins-${{ github.ref }}-${{ github.ref != 'refs/heads/main' || github.run_id }}
       cancel-in-progress: ${{ github.repository != 'cvc5/ethos' || startsWith(github.ref, 'refs/pull/') }}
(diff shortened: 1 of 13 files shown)


daniel-larraz:
Review of `93d584f0`. I built `ethos-eoc` in Debug and ran the vc/sygus/lean/batch pipelines plus cvc5 to confirm these. Eight findings, roughly in severity order.

### 1. `qexp_total` is declared `Int` but axiomatized with Reals — `plugins/smt_meta/smt_meta.smt2:65`

```smt2
(declare-fun qexp_total (Real Int) Int)
(assert (! (forall ((x Real) (y Int))
  (! (= (qexp_total x y) (ite (< y 0) 0.0 (ite (= y 0) 1.0 (* x (qexp_total x (- y 1))))))
  :pattern ((qexp_total x y))))
  :named smtx.qexp_total.def))
```

The body is Real-valued (`0.0`, `1.0`, `(* x ...)` with `x : Real`) while the range is `Int`, so the axiom forces every value to be integral and is contradictory at non-integral bases: `(qexp_total 0.5 1) = (* 0.5 (qexp_total 0.5 0)) = 0.5`, but that term has sort `Int`.

This is a verification hole rather than a crash — a VC that instantiates the pattern at a non-integral base goes `unsat` vacuously and the rule reports "verified" without having been checked. `zexp_total` directly above is consistently `Int`; the range here should be `Real`.

### 2. `char_valid` has no lower bound — `plugins/smt_meta/smt_meta.smt2:56`

```smt2
(define-fun char_valid ((x Int)) Bool (< x 196608))
```

`Char` is `(define-sort Char () Int)`, so negative code points satisfy `char_valid` and are accepted by `$smtx_typeof_value` / `$smtx_value_canonical_bool` as well-typed canonical characters. The Lean model uses `abbrev native_Char := Nat` (`lean_meta_smt_eval.lean:15`) and cannot express those, so a solver can refute a string/char rule VC with `(vsm.Char (- 1))` — a counterexample with no counterpart in the reference semantics. Suggest `(and (<= 0 x) (< x 196608))`.

### 3. Dead prefix test: `$smt_type_` / `$smt_datatype_` do not exist — `plugins/smt_meta/smt_meta_reduce.cpp:439`

```cpp
if (sname.compare(0, 14, "$native_apply_") == 0
    || sname.compare(0, 10, "$smt_type_") == 0
    || sname.compare(0, 14, "$smt_datatype_") == 0)
```

No `$smt_type_*` or `$smt_datatype_*` symbol exists anywhere in the tree. The real names are `$native_type_*` and `$native_datatype`, as used in `meta_reduce_plugin.cpp:117-118` and in the Lean twin at `lean_meta_reduce.cpp:564-565`, which has the correct prefixes.

So the branch is dead for opaque type applications: a term built from `($native_type_0 "UserOp")` (declared at `plugins/desugar/native_embed.eo:17`, used throughout `eo_desugar_native.eo`) skips the embed-name path and falls through to the generic printer, emitting the `?` builtin sentinel — e.g. `(?$native_type_0 (eo.String "Int"))`, which is not valid SMT-LIB.

### 4. Typed non-terminals added twice to `G_eo.Term` — `plugins/smt_meta/smt_meta_sygus.cpp:79`

`finalizeGrammars` appends every entry of `d_grammarTypeAlloc` to the `G_eo.Term` rule list, and then the `d_grefs` loop appends the same non-terminals again for `d_grefs[d_null]`. That loop's `processed` set only dedupes within itself, not against the first loop. In the generated `.sy` all nine `T_*` non-terminals appear twice in `G_eo.Term`, doubling that grammar's branching factor for no benefit. A set shared by both loops would fix it.

### 5. `d_gisFinalized` is never initialized in the constructor — `plugins/smt_meta/smt_meta_sygus.cpp:27`

`SmtMetaSygus::SmtMetaSygus(State& s) : StdPlugin(s) {}` leaves `bool d_gisFinalized` (`smt_meta_sygus.h:115`) indeterminate; it is only assigned in `initializeGrammars`, which is conditional. If `allocateGrammar` runs first, the `Assert(!d_gisFinalized)` at line 112 reads an uninitialized value and can fire spuriously in a Debug build. Worth a `= false` on the member declaration.

### 6. Leftover debug print — `plugins/smt_meta/smt_meta_reduce.cpp:898`

```cpp
std::cout << "ECHO " << msg << std::endl;
```

in `echo()`. Writes stray `ECHO ...` lines to stdout, which corrupts generated output when the plugin prints to stdout.

### 7. `batch --keep-going` does not survive `RuntimeError` — `tools/eoc/driver.py:998`

The loop catches only `subprocess.CalledProcessError`, but `run_vc` raises `RuntimeError` for a rule excluded by the defs file (`driver.py:596`). `--all-rules` pre-filters excluded rules and is unaffected, but `batch --keep-going --rules-file rules.txt` (or explicitly named rules) where one entry is excluded aborts the whole batch — no remaining rule is attempted and no failure summary is printed. Catching `RuntimeError` alongside would restore the intended behaviour.

### 8. `desugar` subcommand silently switched to the VC plugin — `tools/eoc/driver.py:724`

`run_desugar` passes `use_vc_plugin=True`, so the standalone `desugar` subcommand now invokes `--plugin.desugar-vc` instead of `--plugin.desugar`; its `desugar.eo` output changes (VC programs replace `$eo_prog_*`). Every Lean call site was explicitly given `use_vc_plugin=False` to preserve the old behaviour, which makes this one look possibly unintentional. If it is intended, the `### desugar` section at `tools/eoc/README.md:347` should say so.

---

Checked and found OK: the new CI smoke tests (all three output paths exist and are non-empty locally), the paren/nesting accounting in `finalizeProgram` for both the axiom and `define-fun` forms, the curried `FUNCTION_TYPE` encoding in both the pattern matcher and the term printer, forward-declaration and recursion handling, and the generated `.smt2`/`.sy` files parsing and solving under stock cvc5 1.3.4. The `93d584f0` cvc5-discovery change is also fine — `shutil` is imported.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

daniel-larraz (approved):

PR #231, opened 2026-08-27
Add semantics compiler infrastructure

ajreynol:
This is work towards providing a usable frontend to the eoc compiler.  This allows the eoc pipeline to be configurable by a "semantics" configuration files.

In other words 4 files that were previously hardcoded now are compiled via a configuration file / python parser:
- smt_defs.eo / cpc_defs.eo, which are the raw Eunoia that are fed into the model smt layer,
- termination.lean / cpc_termination.lean, which are annotations on the final Lean compilation.

We introduce two *.eos files for this that serve as the ground truth for the current compilation, in tools/eoc/semantics. One for the semantics configuration of SMT (smt.eos). This is intended to live here. The other is development-cpc.eos, which is used for testing (this file will eventually live in the Logos repo).

This also refactors the hardcoded model_smt.eo template to work more of the SMT semantics into configuration, making things more agile.


daniel-larraz (commented):
Review of the semantics compiler infrastructure. I built `ethos-eoc`, ran the compiler, ran the CI smoke test end to end (passes), and diffed the generated `smt_defs.eo`/`user_defs.eo` against the deleted checked-in `smt_defs.eo`/`cpc_defs.eo` — the differences are alias/inlining rewrites plus intended configuration changes, and I found no semantic regression in the emitted model cases.

Seven findings below, each reproduced. The theme of the first three is that validation added alongside the compiler does not currently run.

No correctness problem found in `rtrimLines` itself, in the new `d_keep`/`$emb_tsm.` handling in `defs_reader.cpp`, or in the new type-constructor/case emission ordering in `model_smt.cpp` — all type blocks are `eoc-keep`, so the configuration order is stable as claimed.

tools/eoc/sem_compile.py
```diff
@@ -0,0 +1,622 @@
...
+    self.vocab = vocab
+    self.macros = macros
+    self.decls = decls
+    # Whether a name in quotes the embedding does not have is an operator of
+    # the value layer or a misspelling, which is read off whether the set has a
+    # value layer at all. See Decls.raw_operators.
+    self.raw_operators = decls.raw_operators
+    self.missing = []
+
+  def need(self, name, who, why='names'):
```
daniel-larraz, line 212: **`Ctx.need()` is never called, so this validation is dead.**

Nothing in the tree calls `need()`, so `self.missing` stays empty and `Ctx.check()` always passes. The documented guarantee that every helper a body names is written out by some file of the set is not enforced.

Reproduced: adding a body that names `$smtx_totally_bogus_helper` compiles silently into `smt_defs.eo` with no diagnostic.

This also leaves `Ctx.written`, `defined_names()`, `Block.defines()` and `Symbol.defines()` with no live consumer.

tools/eoc/sem_compile.py
```diff
@@ -0,0 +1,622 @@
...
+  a = ap.parse_args()
+  paths = a.configs or list(CONFIGS)
+  global ALIASES
+  ALIASES = build_aliases(read_vocabulary(VOCAB_FILES),
+                          read_macros(MACRO_FILES))
+  if a.check:
+    bad = 0
+    for config, blocks in compile_all(paths):
+      name = os.path.basename(config.target)
+      bad += check(blocks, config.target, a.verbose)
```
daniel-larraz: **`--check` compares the output against itself, and crashes on a fresh clone.**

`check(blocks, config.target, ...)` and `check_forms(blocks, config.target)` read `config.target` — the very file `main()` writes a few lines below. So the comparison is between the compiled blocks and the last thing the compiler wrote, which can never differ once the compiler has run.

On a fresh clone it is worse: `tools/eoc/out` is gitignored and the old checked-in reference files (`smt_defs.eo`, `cpc_defs.eo`) are deleted in this PR, so `config.target` does not exist and `--check` dies with an uncaught `FileNotFoundError`.

Only `check_order` still tests a real property. If the intent is "the checked-in reference matches what the compiler produces", the reference needs to be a separate committed file; otherwise consider dropping the comparison so `--check` does not read as coverage it does not have.

tools/eoc/sem_compile.py
```diff
@@ -0,0 +1,622 @@
...
+  back what each set compiled to, so the caller need not know the layout.
+  """
+  out = {}
+  for config, blocks in compile_all(paths):
+    target, lean_target = config.target, config.lean_target
+    if out_dir is not None:
+      target = os.path.join(out_dir, os.path.basename(target))
+      lean_target = os.path.join(out_dir, os.path.basename(lean_target))
+    write_if_changed(render(blocks, config), target)
+    write_if_changed(render_lean(config), lean_target)
```
daniel-larraz, line 462: **The generated `*_termination.lean` files are written but never read.**

`render_lean` emits `smt_termination.lean` / `user_termination.lean`, but `LeanMetaReduce` hard-codes `plugins/lean_meta/termination.lean`, and `--lean-config` defaults to `None`, so nothing loads the generated files. They duplicate the hand-written ones clause-for-clause today, which hides the disconnect.

Consequence: editing a `:lean` attribute in a `.eos` — the place the docs point at — has no effect on the Lean backend. Either wire `LeanMetaReduce` to the generated file or drop the emission until it is wired.

tools/eoc/sem_lang.py
```diff
@@ -0,0 +1,1504 @@
...
+      cases.append((' '.join(cast(p, Scope(bound=names), self, ctx, 'value')
+                             for p in pats.items),
+                    cast(body, Scope(bound=names), self, ctx, agg.level)))
+    return agg.render_helper(self, cases, params, ctx, len(self.params), total)
+
+
+# -- reading one --------------------------------------------------------------
+
+
+def _places(pat, declared, agg):
```
daniel-larraz: **`_places` allocates slots without the argument kind, so explicit patterns can emit undeclared names and unrewritten heads.**

`_places` walks `pat.raw` and allocates slots positionally, with no notion of the argument kind, while `Aggregate.params` declares parameters *by kind*. The head is then emitted straight from the raw pattern rather than from the rewritten symbol.

Reproduced: a `define-sort` case with an explicit pattern emits

```
(($eoc_type_default_BitVec (BitVec x1)) ...)
```

under a program that declares `(n1 $native_Nat)` — `x1` is never declared, and `BitVec` was not rewritten to `$tsm_BitVec`. No diagnostic is produced for either.

plugins/model_smt/model_smt.cpp
```diff
@@ -68,14 +73,22 @@ void ModelSmt::finalizeDecl(const std::string& name)
 
 void ModelSmt::loadDefs()
 {
-  if (!d_smtDefs.read(getResourcePath("plugins/model_smt/smt_defs.eo")))
+  // The SMT-LIB signature is the target of the compilation, so the plugin
+  // ships with one and reads that where it was given no other; --semantics is
+  // what gives it another, the way --signature gives it the input's.
+  const std::string smtDefs =
+      d_smtDefsFile.empty() ? getResourcePath("tools/eoc/out/smt_defs.eo")
```
daniel-larraz, line 83: **The "bundled" SMT-LIB semantics is a generated, gitignored file.**

The fallback resource is `tools/eoc/out/smt_defs.eo`, but `tools/eoc/out` is gitignored and this PR deletes the checked-in `smt_defs.eo`. So the plugin only "ships with one" after `sem_compile.py` has been run locally.

Reproduced on a fresh clone + build: the documented standalone invocation

```
ethos-eoc --plugin.model-smt --signature=... <file>
```

fatals here until the compiler has been run. Either install the generated file as a build artifact (so the resource path resolves after `make`), or keep a committed copy as the fallback.

tools/eoc/sem_compile.py
```diff
@@ -0,0 +1,622 @@
...
+    return name_of(self.path)
+
+  @property
+  def is_target(self):
+    """Whether the set is the SMT-LIB signature, which is the target of the
+    compilation, rather than the signature of an input.
+
+    Which one a set is is said by what it is called, since the two compile to
+    different things and a run has to know which before it reads a line.
+    """
```
daniel-larraz: **Compile shape is keyed on basename, and non-shipped sets collide on one output path.**

`is_target` decides the *entire* compile shape from `self.name == SMT_SET`, i.e. from the basename alone. Combined with `_beside` (below), which writes any set not in `CONFIGS` next to itself, this gives two silent failures for out-of-tree configurations, both reachable from `compile_signatures` in `driver.py`:

1. `--semantics mytree/mysmt.eos` has basename `mysmt`, so it is compiled with the **input** shape rather than the target shape — wrong output, no error.
2. Two non-`smt` configs in the same directory both resolve to `<dir>/user_defs.eo`. The stage can then be handed the same file as both `--signature` and `--semantics`.

Keying the shape on something explicit in the `.eos` itself, rather than on the filename, would avoid both. At minimum the second case deserves a collision check.

plugins/utils.h
```diff
@@ -101,6 +101,16 @@ bool isEmbedMetaKind(MetaKind k);
  */
 const std::string& getParseDefPrefix();
 /** Return true if name is the name of a parse definition. */
+/**
```
daniel-larraz: **This declaration was inserted between `isParseDefName`'s doc comment and `isParseDefName`.**

Line 103 (`/** Return true if name is the name of a parse definition. */`) documents `isParseDefName`, which now sits at line 114 — the new `rtrimLines` comment and declaration landed in between, so the comment reads as documenting `rtrimLines`.

Moving the `rtrimLines` block above line 103 (or below `isParseDefName`) restores the pairing.

daniel-larraz (approved):

PR #232, opened 2026-08-27
Miscellaneous changes from eoc branch

ajreynol:
Includes whitespace changes.

### DIFF
--- a/plugins/lean_meta/lean_meta_checker.lean
+++ b/plugins/lean_meta/lean_meta_checker.lean
@@ -142,10 +142,4 @@ $LEAN_CHECKER_DEFS$
 inductive eo_is_refutation : Term -> CCmdList -> Prop
 $LEAN_EO_IS_REFUTATION_DEF$
 
-/-- API for logos -/
-def logos_init_state : CState := CState.nil
-def logos_invoke_assume (s : CState) (A : Term) : CState := (CState.cons (CStateObj.assume A) s)
-def logos_invoke_cmd (s : CState) (c :CCmd) : CState := (__eo_invoke_cmd s c)
-def logos_state_is_refutation (s : CState) : native_Bool := (__eo_state_is_refutation s)
-
 end Eo
--- a/plugins/lean_meta/lean_meta_reduce.cpp
+++ b/plugins/lean_meta/lean_meta_reduce.cpp
@@ -663,9 +663,6 @@ void LeanMetaReduce::defineProgram(const Expr& v, const Expr& prog)
 
 void LeanMetaReduce::finalizePrograms()
 {
-  std::set<Expr> progProcessed;
-  std::vector<Expr> waiting;
-  std::set<Expr> waitingDef;
   for (size_t i = 0, nprogs = d_progDefs.size(); i < nprogs; i++)
   {
     Expr prog = d_progDefs[i];
@@ -673,7 +670,6 @@ void LeanMetaReduce::finalizePrograms()
     Expr def = d_progToDef[prog];
     finalizeProgram(prog, def, isDefine);
   }
-  Assert(waiting.empty());
 }
 
 void LeanMetaReduce::finalizeProgram(const Expr& v,
--- a/plugins/lean_meta/lean_meta_reduce.cpp
+++ b/plugins/lean_meta/lean_meta_reduce.cpp
@@ -328,7 +328,7 @@ void LeanMetaReduce::printEmbAtomicTerm(const Expr& c, std::ostream& os)
       os << bv.getSize() << " " << bvi.toString() << ")";
     }
     else if (k == Kind::STRING)
-    {     
+    {
       os << "(Term.String ";
       std::string css = l->toString();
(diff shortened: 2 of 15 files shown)


plugins/lean_meta/lean_meta_checker.lean
```diff
@@ -142,10 +142,4 @@ $LEAN_CHECKER_DEFS$
 inductive eo_is_refutation : Term -> CCmdList -> Prop
 $LEAN_EO_IS_REFUTATION_DEF$
 
-/-- API for logos -/
-def logos_init_state : CState := CState.nil
```
daniel-larraz, line 146: These four definitions were added in #228 and are removed here with no replacement. Nothing in this repo references them, but `driver.py`'s `LEAN_OUTPUTS` publishes the generated form of this template as `Logos.lean` into the downstream Lean package, and the block was explicitly labelled `/-- API for logos -/`. If that package calls `logos_init_state` / `logos_invoke_assume` / `logos_invoke_cmd` / `logos_state_is_refutation`, this breaks it silently from here.

Is the removal intentional and coordinated with the downstream package?

daniel-larraz (commented):
Reviewed the diff. Most of it is whitespace cleanup of the `.eo`/`.lean`/`.smt2` templates plus three new `rtrimLines()` call sites; the only substantive logic change is the option-parsing rewrite in `main_eoc.cpp`. Inline comments below.

**Checked and clean:**

- `rtrimLines` boundary behaviour is correct (empty input, trailing newline, `pos == size()`), and all four output paths now trim consistently — no `.eo`/`.lean`/`.smt2` resource under `plugins/` still carries trailing whitespace, so the verbatim `copyResourceToOutput` path stays in sync. (It doesn't strip `\r`, which only matters if CRLF templates ever appear.)
- The reordering of the option block is safe: `--include=` / `--reference=` are still matched *before* the new `--` catch-all, so they don't get swallowed by it, and the same holds for `--no-parser`, `--signature=`, `--semantics=`, `--lean-config=`. `compare(0, n, ...)` can't throw at `pos == 0`, and the `--no-` / `--` prefix split computes `val` and `key` correctly.
- Dropping the post-parse `d_requireProofOfFalse` check loses no coverage: it defaults to `false`, so only the command line could set it, and that is now caught inline.
- `-t`'s tag argument is still consumed correctly, and `--stats` still reaches `setOption` and takes effect at the end of `main`.
- `git diff -w` confirms the template edits are whitespace-only. The `../utils.h` includes match the existing convention in `plugins/`.

plugins/lean_meta/lean_meta_reduce.cpp
```diff
@@ -672,68 +672,6 @@ void LeanMetaReduce::finalizePrograms()
...
-          }
-          if (waiting.size() > 1)
-          {
-            d_defs << "end" << std::endl;
-          }
-          waiting.clear();
-        }
-    */
   }
   Assert(waiting.empty());
```
daniel-larraz: With the commented-out mutual-block minimization gone, the three locals declared at the top of `finalizePrograms` (`progProcessed`, `waiting`, `waitingDef`, lines 666-668) are never written, so this `Assert` is vacuous. In a release build `Assert` expands to `EO_FATAL_IF(false, ...)`, which leaves `waiting` unreferenced entirely.

Suggest dropping all four lines.

plugins/main_eoc.cpp
```diff
@@ -171,6 +169,25 @@ int main(int argc, char* argv[])
...
+    {
+      bool val = (arg.compare(0, 5, "--no-") != 0);
+      std::string key = arg.substr(val ? 2 : 5);
+      if (isUnsupportedOption(key))
+      {
+        EO_FATAL() << "Error: " << arg << " is not supported by ethos-eoc";
+      }
+      if (!opts.setOption(key, val))
+      {
+        EO_FATAL() << "Error: unrecognized option " << arg;
```
daniel-larraz, line 189: `--help` and `--show-config` are real ethos options (handled in `src/main.cpp`), just not wired into `ethos-eoc`, so this catch-all now reports them as *unrecognized*. Not a regression — the old code fell through and treated `--help` as the input filename, which was worse — but since this block is being rewritten anyway, the sibling wording a few lines up (`is not supported by ethos-eoc`) reads better here, or `--help` could just be forwarded so `ethos-eoc --help` prints usage.

tools/eoc/README.md
```diff
@@ -402,6 +402,19 @@ This walks `include` chains and preserves declaration order.
 
 ## Common workflows
 
+### Generate a VC for one rule
```
daniel-larraz: Minor: both of these examples already appear further down in this same section (the `vc` one and the two `batch` ones). Might be worth removing the duplicates rather than adding a second copy above them.

plugins/lean_meta/lean_meta_checker.lean
ajreynol: Yes it is coordinated.

daniel-larraz (approved):

PR #233, opened 2026-08-28
Fixes for reference checking

ajreynol:
This ensures we do proper reference checking based on the commands `check-sat-assuming` and `reset-assertions`.

This also makes it so that we uniformly ignore all `(get-...)` commands in reference files, independent of the current SMT-LIB set of commands.

The user manual is updated to document this.


daniel-larraz (commented):
I built this branch and ran the suite: 188/188 tests pass, no new compiler warnings, and the three new fail-tests fail for the right reasons.

The change is well-scoped and correct in what it does. The `get-` prefix dispatch, the s-expression skipping loop, and `clearReferenceAsserts` all behave properly on the edge cases I probed: `(get-value ())`, `(get-info :status)`, string literals containing unbalanced parens `(get-model :foo "a )( b" 3.5 #b101)`, non-2.6 names like `(get-difficulty)`, and truncated input `(get-model` -> clean `Expected s-expression` error. `parseSymbolicExpr`'s `nparen--` underflow is correctly avoided by the `RPAREN` peek guard in `Token::SMT2_QUERY_COMMAND`, and the trailing `d_lex.eatToken(RPAREN)` still consumes the command's closing paren. In `clearReferenceAsserts`, clearing the raw-pointer set *before* the ref-holding list is the right order; reversing it would risk stale pointers.

Three things I think are worth addressing.

### 1. `push`/`pop` remain silently unsound, by the same rationale that motivated the `define-fun-rec` error

The justification given for rejecting `define-fun-rec` is that ignoring it would silently drop assertions from the query. `pop` has the worse failure mode: it silently *retains* assertions that the query discarded. Concretely, with a reference file

```smt2
(set-logic ALL)
(declare-fun x () Int)
(push 1)
(assert (< x 0))
(check-sat)
(pop 1)
(assert (> x 0))
(check-sat)
```

and a proof containing `(assume @p0 (< x 0))`, ethos prints `correct`. The proof assumes a formula that is not part of the final query.

This is pre-existing behavior, but this PR now documents it ("Reference checking should not be run ... on benchmarks that contain `push` or `pop`") without enforcing it, while enforcing the analogous `define-fun-rec` case and implementing the analogous `reset-assertions` case. Two options: error out on `push`/`pop` when `d_isReference` (`src/cmd_parser.cpp`, `case Token::POP`/`Token::PUSH`), or scope the reference assertions properly (record `d_referenceAssertList.size()` on push, truncate and rebuild the set on pop). The latter is only a few lines and would also provide the machinery to scope `check-sat-assuming` assumptions later.

### 2. `(reset)` does not clear the reference assertions

`case Token::RESET` calls `State::reset()`, which does not touch `d_referenceAsserts`. In SMT-LIB, `reset` subsumes `reset-assertions`, so it should also call `clearReferenceAsserts()`. Impact is low in practice, since `State::reset()` clears the shared `d_symTable` and a reference file containing `reset` therefore errors out on the next symbol anyway (I checked several variants, all abort), but it is a one-line inconsistency now that `reset-assertions` is handled.

### 3. `check-sat-assuming` skips the Bool type check that `assert` performs

`Token::ASSERT` uses `parseFormula()`, which type-checks the term against Bool. The new `Token::CHECK_SAT_ASSUMING` case uses `parseExprList()`, which does not. As a result `(check-sat-assuming (x))` with `x : Int` is accepted silently and inserts a non-Bool term into the reference assertion set.

This is not exploitable, since `assume` type-checks its own formula and so the bad entry can never be matched, but it lets malformed input through where `assert` would catch it. A `typeCheck(a, d_state.mkBoolType())` inside the loop would restore parity.

### Minor

- **Doc**: "Any command not listed in the grammar below is not recognized and leads to a parse error" is not quite true. `push` and `pop` are accepted but are deliberately absent from the grammar (they are marked "undocumented" in the command table). Worth softening, or adding them once point 1 is resolved.
- **Test gap**: `define-funs-rec` has no regression test, only `define-fun-rec`. They share a `case` label so the risk is low, but it is one small file.
- `str.compare(0, 4, "get-")` is safe for symbols shorter than 4 characters (`pos == 0 <= size()`, so no `out_of_range`). No change needed; noting it only because it is the kind of thing that reads as risky.

ajreynol:
We do not yet support push/pop by design. We will not add this feature on this PR.

ajreynol:
Also I don't think its worthwhile to hardcode define-fun-rec into the error handling.

daniel-larraz (approved):

PR #234, opened 2026-08-29
More fixes for reference checking

ajreynol:


src/expr_parser.cpp
```diff
@@ -195,6 +218,35 @@ Expr ExprParser::parseExpr()
...
+              {
+                d_lex.nextToken();
+                uint32_t w = parseIntegerNumeral();
+                d_lex.eatToken(Token::RPAREN);
+                BitVector bv(w, Integer(bvval));
+                ret = d_state.mkLiteral(Kind::BINARY, bv.toString());
+              }
+              else
+              {
+                pstack.emplace_back(ParseCtx::NEXT_ARG);
```
daniel-larraz: In reference mode, `(_)` now crashes instead of giving a parse error. After `_` is dropped, the argument list is empty, so if `)` comes next the parser calls `mkExpr(Kind::APPLY, {})` with no arguments. A debug build fails an assert. A release build reads `vchildren[0]` of an empty vector, which is undefined behavior and likely a segfault.

Inputs that trigger it: `(assert (_))` or `((_) x)` in a reference file.

Suggested fix: return a parse error here when `_` is followed directly by `)`, for example by checking `d_lex.peekToken() == Token::RPAREN` before pushing `NEXT_ARG`.

ajreynol:
The issue referenced by the review is fixed on https://github.com/cvc5/ethos/pull/245 I believe.

daniel-larraz:
@ajreynol I built this PR merged with #245 (debug build), and `(_)` in a reference file still crashes. #245's check only matches when `_` is the single collected argument, which is the Eunoia case. In reference mode this PR drops `_` and pushes a frame with no arguments, so `sf.d_args` is empty when `)` is reached:

| Input | #234 + #245 |
|---|---|
| reference file: `(assert (_))` | `Check failure !children.empty()` in `State::mkExpr` |
| reference file: `(assert ((_) x))` | `Check failure !children.empty()` in `State::mkExpr` |
| Eunoia: `(define y () (_))` | `Error: Expected an operator for _` |

Adding a check for empty arguments fixes both reference cases. With it, both give `Expected an operator for _`:

```cpp
if (sf.d_args.empty()
    || (sf.d_args.size() == 1 && sf.d_args[0] == d_state.getVar("_")))
{
  d_lex.parseError("Expected an operator for _");
}
```

A reference-file version of the `apply-no-operator` regression test would cover this path.

ajreynol:
@daniel-larraz let's review/merge #245 first. WDYT?

daniel-larraz (approved):

PR #235, opened 2026-08-31
Major updates to the agility of the eoc compiler

ajreynol:
Major progress here in terms of agility: hardcoded *.eo, *.smt2, *.lean is now captured in *.eos. This is work towards allowing us to formally reason about the compiler.

It also adds usability options for upstream tools.

This makes no change to core ethos, only the eoc compiler.


daniel-larraz:
## Code review

Reviewed the full `main...HEAD` diff: the `NativeLayer` / aggregate-driven `defs_reader` + `model_smt` rewrite, the `LeanMetaReduce` UserOp ladder, the `desugar` UOp generation, and the Python compiler side (`sem_compile.py`, `sem_lang.py`, `sem_target.py`, `report.py`).

What checks out: `sem_compile.py` compiles cleanly against the new configs, both native layers come out topologically ordered, every `$…$` hole in `model_smt.eo` is claimed by exactly one aggregate/helper/embed-datatype/manual marker, and `NativeLayer::use`'s promotion/cycle logic terminates correctly.

Four findings below. The first three are all one root cause — `tools/eoc/driver.py` did not move with this change — and the comment at `.github/workflows/main.yml:54` says the pipeline smoke tests were disabled because driver.py "arrives with the change that adds the cpc wrappers and the regression". So the split looks deliberate; flagging so it isn't lost, since as it stands the tree can't run the pipeline end to end. **Finding 4 is independent of that split** and is the one I'd fix in this PR.

### 1. `tools/eoc/driver.py:509` — stale CLI flags

```python
args.append(f"--signature={self.binary_path_arg(self.defs_file)}")
```

`--signature=` was removed by the `main_eoc.cpp` option rename, so `ethos-eoc` dies with `Error: unrecognized option --signature=…`. Compounding it, `--semantics=` now binds the *input's* semantics, so the SMT-LIB file on the next line goes into the wrong slot.

### 2. `tools/eoc/driver.py:47` — wrong Lean output name

```python
("lean_meta/lean_meta_smt_eval.lean", "SmtEval.lean", False),
```

`finalize()` now renders this file to `lean_meta_smt_eval_gen.lean`, but `LEAN_OUTPUTS` still copies the ungenerated source (`generated=False`). The published `SmtEval.lean` therefore keeps the literal `$NATIVE_DEFS$` and none of the native layer, which breaks every generated Lean module.

### 3. `plugins/desugar/eo_desugar.eo:504` — includes with nothing to resolve them

```
(include "user_desugar.eo")
```

and `(include "native_defs.eo")` at `plugins/desugar/native_embed.eo:70`. `sem_compile.py` writes both files to `tools/eoc/out/`, and the surrounding comment says "which the driver puts here" — but nothing inlines them or places them beside the generated `.eo`, so the next stage aborts on a missing include.

### 4. `plugins/lean_meta/lean_meta_reduce.cpp:1199` — `d_scope` not set in `finalizeDecl`

```cpp
if (printMetaType(dty, tn, tk))
```

This call chain is `finalizeDecl` → `printMetaType` → `getEmbedName` → `d_natives.use(name, d_scope)`, and `finalizeDecl` never sets `d_scope`. A native named by a datatype constructor (e.g. `native_Nat` under `$emb_tsm.BitVec`) is then attributed to whatever module the last program left in `d_scope`. `SmtMetaReduce::finalizeDecl` sets `d_scope` for exactly this reason; the asymmetry can strand a non-`:keep` native in `Logos.lean` (surfacing as an unknown identifier in `SmtModelDefs.lean`) or trigger a spurious `EO_FATAL` out of `NativeLayer::use`.

---

ajreynol:
The first 3 issues reported are all part of the PR plan; they will be addressed in a followup PR

daniel-larraz (approved):

PR #236, opened 2026-09-01
Finalize eoc compiler

ajreynol:
A followup PR will finalize an initial draft of our AI policy FYI @daniel-larraz


daniel-larraz:
Reviewed the branch. The substance here is `tools/eoc/driver.py` — the option rename, the head-line-based `defs_head`/`defs_excludes`/`defs_depends` refactor, `inline_called_blocks`, `--natives=eo`, `--calc-name`, and the `Proofs/` Lean layout — plus the new digest regression in CI.

**What I verified holds:**

- The `defs_excludes`/`defs_depends` refactor is behaviour-preserving: I ran the old and new algorithms side by side against the generated `tools/eoc/out/user_defs.eo` and got byte-identical output (5 excludes, 179 depends lines).
- The `LEAN_ALL_DEPS`/`LEAN_SINGLE_DEPS`/`DESUGAR_VC_DEPS` regroupings are set-identical to the originals, modulo the `$smtx_msm_lookup` → `$smtx_map_lookup` and `$smtx_value_canonical_bool` → `$smtx_value_canonical` renames, which match the symbols actually present in `tools/eoc/semantics/smt.eos`.
- All nine `LEAN_OUTPUTS` sources are emitted by `lean_meta_reduce.cpp`, and the include chain `eo_desugar.eo → eo_desugar_native.eo → native_embed.eo → native_defs.eo` has exactly one marker each, so the inlining order is sound.

**Findings:**

**1. `inline_called_blocks` can splice in undeclared symbols** — [`driver.py:219`](https://github.com/cvc5/ethos/blob/09ce25ec75097bdef6f947c59cc91f50117a24e8/tools/eoc/driver.py#L219)

The function keeps a nil-predicate block based only on whether its own program is applied, never on the symbols the block's *body* names. `$eo_is_list_nil_str.++` names `seq.empty` and `$eo_is_list_nil_@@TypedList.cons` names `@@TypedList.nil`. Since `sem_compile.head_lines` builds `$eoc-depends` from the model blocks only — never from `config.desugar_blocks` — a trimmed run can splice in a block naming something nothing declares. That is the exact failure the docstring says it prevents.

**2. The two-roles guard in `compile_signatures` is bypassable** — [`driver.py:849`](https://github.com/cvc5/ethos/blob/09ce25ec75097bdef6f947c59cc91f50117a24e8/tools/eoc/driver.py#L849)

The guard only fires for shipped sets, since `role_of` returns `None` otherwise, so `--semantics my.eos --smt-semantics my.eos` is accepted. `written` is keyed by `config.path`, so the input-role entry overwrites the target-role one and the driver silently passes `user_defs.eo` as `--smt-semantics`.

**3. CI never exercises the new inlining mechanism** — [`test/regress.py:63`](https://github.com/cvc5/ethos/blob/09ce25ec75097bdef6f947c59cc91f50117a24e8/tools/eoc/test/regress.py#L63)

`tests/Booleans-rules.eo` contains only ground `:right-assoc-nil` terms, so `desugar.cpp` never emits an `($eo_is_list_nil_<sym> nil)` call. `inline_called_blocks` — the central new mechanism in this PR — only ever runs its "keep nothing" branch in CI. A test with a non-ground `:right-assoc-nil` application would exercise the keep branch, and would likely surface finding 1 directly.

**4. Destructive step precedes the fallible one** — [`driver.py:631`](https://github.com/cvc5/ethos/blob/09ce25ec75097bdef6f947c59cc91f50117a24e8/tools/eoc/driver.py#L631)

`lean()` rmtrees the published tree before invoking ethos, so a failing lean-meta run leaves an empty `lean/` where the last good tree was. Moving the wipe to after the ethos call keeps the intent without the window.

daniel-larraz (approved):

PR #243, opened 2026-09-19
Remove deprecated scripts, update to standard layout

ajreynol:
The eunoia formatter now lives in another repo (https://github.com/ajreynol/eunoia).

This PR also updates the standard recommended layout, based on https://github.com/ajreynol/kanon.

Also removes a description of *.eos files, which is now maintained by https://github.com/ajreynol/eunoia as well.


daniel-larraz:
## Review notes

Two things came up; one is a definite fix.

### Stale links after the compiler move — `tools/eoc/docs/design.md:45`

The three table links still point at the pre-move paths `../sem_lang.py`, `../sem_target.py`, `../sem_compile.py`. All three files now live in `tools/eoc/compiler/`, so those links 404. The prose in the same file (lines 4-6 and 55) was updated, so this looks like an oversight rather than intent. Should be `../compiler/sem_lang.py` and so on.

### Out-of-tree callers of `sem_compile.py` (worth confirming, not necessarily a problem)

`report.py`'s docstring says logos and cvc5 CI run this compiler. Every in-repo caller was updated for the move, but anything in another tree that names `tools/eoc/sem_compile.py` by path would break. The new `tools/eoc/README.md` asserts `driver.py` is "the one path a caller of another tree names" — if that holds against the actual CI configs, there's nothing to do here.

### Checked and clean

- No dangling references anywhere to the removed `contrib/` scripts (`eo_format.py`, `alfc_*`, `drat_trim`) or to the removed `tools/eoc/docs/semantics.md`.
- `sys.path` plumbing is correct in all four places after the move: `driver.py` (module level and `compile_signatures`), `sem_compile.py`, `test/regress.py`, and `report.ROOT` — the four `dirname` calls do resolve to the repo root.
- The module-level ordering in `driver.py` is right: `compiler/` is inserted first and the driver's own directory second, so the flat copy-out used by `test/driver.py` still finds `report.py` beside the driver.
- Built `ethos-eoc`, deleted `tools/eoc/out/`, and rebuilt through the CMake `eoc-semantics` target — the reflowed multi-line `COMMAND` in `plugins/CMakeLists.txt` works.
- `python3 tools/eoc/compiler/sem_compile.py --check` — all 9 generated files current.
- `python3 tools/eoc/test/driver.py` — 3 tests OK.
- `python3 tools/eoc/test/regress.py` with cvc5 read-back — "42 files, all as checked in", so the updated `expected.txt` hashes are correct.
- The new external doc URL returns 200 and the `#5-the-shape-of-what-is-written` anchor matches a real heading.

Two nits that predate this PR, noted only in passing: `tools/eoc/compiler/sem_compile.py:18` references a nonexistent `sem_decl.py`, and `.gitignore:12` references a nonexistent `install_logos`.

---

*This review was AI assisted (Claude Code).*

daniel-larraz (approved):

